PR #4, opened 2011-10-29
Support skipping on user triggered builds

petehayes:
I'd like to be able to prevent Sonar from running when there is a forced build by a user, even when there is no SCM changes.  My users sometimes do this to run a full build of all modules instead of an incremental build.


henryju:
Now that pipeline is the way to go, I'm closing this as won't fix.

PR #5, opened 2011-12-07
SONARPLUGINS-1549: Sonar "aggregator" mode

henryju:
Here is a proposal to have "aggregator" mode in Jenkins for Sonar analysis


henryju:
Outdated

PR #6, opened 2012-04-12
Added lookup for skip envVar in the environment to allow adding the skip envVar via the Environment Inject plugin.

akuhtz:
Added lookup for skip envVar in the environment to allow adding the skip envVar via the Environment Inject plugin.


dbmeneses:
This was already introduced in dc4e883ecfe00f39c5e74b77cc564f7403bee2d6 - closing it.

PR #9, opened 2013-12-19
Store passwords as secrets

olivergondza:


jglick:
Unclear which of this and [SONARJNKNS](http://jira.codehaus.org/browse/SONARJNKNS) is actually used. Clearly one or the other should be closed for entries.

henryju:
Manually merged. Thanks again for your contribution.

PR #10, opened 2014-04-08
Fix collision of jdk configuration between the main job configuration and a Sonar task

ydubreuil:
This pull-request fixes a bug which corrupts the job jdk field. The jdk field is overwritten with the value of the jdk of the sonar task which defaults to 'Hérite du job parent' in french. This breaks the build.

Fix is straightforward, it just adds the plugin name as a prefix for the jdk field in the jelly file.

Thanks for merging,
Yoann.


henryju:
Seems this is no more an issue with recent versions of Jenkins since form parameters are serialized as JSON object with proper namespace. But for "purity" I'll merge this P/R.

PR #13, opened 2014-12-09
Update dependencies and fix tests

jirutka:


henryju:
Manually applied (part of http://jira.codehaus.org/browse/SONARJNKNS-206). Thanks for your contribution.

PR #15, opened 2015-03-09
Sonarqube badge should be a link to the project page

or-shachar:
I love the little sonarqube badge that is added to the run history but I think it must be made more functional by making it a link to the SonarQube project.
Do you think it can be problematic?


henryju:
Ticket created http://jira.codehaus.org/browse/SONARJNKNS-216. Will be considered during next sprint. Thanks for your contribution.

or-shachar:
well? :)

dbmeneses:
The proposed change does not work, but I'll implement it in this sprint. Thanks for the idea.

PR #16, opened 2015-04-02
Prevent loading all builds to generate the url

christ66:
In order to generate the url at the root project page it could be an expensive operation because it will iterate all of the builds in order to find a url. If a project has a lot of failed/aborted builds which never sent information to sonar (and a `BuildSonarAction` was never created), or if the project enabled sonar for the first time would most likely encounter a performance issue and load every build until it found a sonar url. Instead this pull request changes it to generate the url from the last stable or unstable build.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
+++ b/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
@@ -39,7 +39,6 @@ import hudson.model.Run;
 import hudson.plugins.sonar.BuildSonarAction;
 import org.apache.commons.io.IOUtils;
 
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -98,7 +97,11 @@ public final class SonarUtils {
   }
 
   @Nullable
-  private static String getSonarUrlFromRun(@Nonnull Run run) {
+  private static String getSonarUrlFromRun(Run run) {
+    if (run == null) {
+      return null;
+    }
+
     BuildSonarAction action = run.getAction(BuildSonarAction.class);
     if (action != null) {
       return action.getUrlName();
--- a/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
+++ b/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
@@ -39,7 +39,7 @@ import hudson.model.Run;
 import hudson.plugins.sonar.BuildSonarAction;
 import org.apache.commons.io.IOUtils;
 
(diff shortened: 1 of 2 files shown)


ndeloof:
:+1:

src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
```diff
@@ -88,13 +88,25 @@ public static String extractSonarProjectURLFromLogs(AbstractBuild<?, ?> build) t
...
+    String url = getSonarUrlFromRun(project.getLastSuccessfulBuild());
+    if (url != null) {
+      return url;
+    } else {
+      return getSonarUrlFromRun(project.getLastUnstableBuild());
+    }
+  }
+
+  @Nullable
+  private static String getSonarUrlFromRun(Run run) {
```
jglick: Use `Run<?,?>` to avoid an `unchecked` warning.

src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
```diff
@@ -88,13 +88,25 @@ public static String extractSonarProjectURLFromLogs(AbstractBuild<?, ?> build) t
...
-      }
+    String url = getSonarUrlFromRun(project.getLastSuccessfulBuild());
+    if (url != null) {
+      return url;
+    } else {
+      return getSonarUrlFromRun(project.getLastUnstableBuild());
+    }
+  }
+
+  @Nullable
```
jglick: Use `@CheckForNull`. In general avoid `@Nullable` unless you have read its Javadoc carefully and really know what you are doing.

src/test/java/hudson/plugins/sonar/utils/SonarUtilsTest.java
```diff
@@ -67,7 +67,7 @@ public void shouldGetLastUrl() throws Exception {
     RunList list = mock(RunList.class);
     when(list.iterator()).thenReturn(Arrays.asList(build1, build2).iterator());
     when(project.getBuilds()).thenReturn(list);
-    assertThat(SonarUtils.getLastSonarUrl(project)).isEqualTo("http://foo");
+    assertThat(SonarUtils.getLastSonarUrl(project)).isNull();
```
jglick: Would be better to ditch the `RunList` and mock `getLastSuccessfulBuild`.

christ66:
@henryju ping?

henryju:
Sorry @christ66 I missed your P/R. It is merged and I will try to release is as part of upcoming 2.2.1. Thanks for your contribution.

PR #18, opened 2015-09-24
SONARJNKNS-213: Jenkins Workflow support

vaskrist:
I tried to make minimal changes, maintain the API, and not add to the dependencies, except for the workflow dependency for testing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,12 @@
       <version>1.7.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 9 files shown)


henryju:
The code change look good to me but I will probably not have time to do extensive testing. Are you using this code in production?

vaskrist:
I am using it in a proof of concept with workflow right now. What I didn't try yet is to use something to break the build like the build breaker plugin and see if it breaks the workflow build too. Let me do that tomorrow and I will come back to you with a positive reply or an updated pull request.

dbmeneses:
@vaskrist did you end up doing the test?

dbmeneses:
This LGTM. I didn't test breaking the build yet but if necessary we'll work on top of it.

PR #28, opened 2015-11-06
Fix sonar-jdk in post-build action (publisher)

mvanini:
On edit config job, sonar-jdk in config.jelly doesn't match with actual sonar-jdk. On submit it will overwrite previous value with null (inherit from job), causing misconfiguration in job build.


mvanini:
Why travis's 4th sub-build is broken?

dbmeneses:
There is a problem with the P/R analysis script, but the P/R is otherwise ok.

henryju:
@mvanini  Please rebase your P/R on master, it should fix Travis build (no P/R analysis is possible for external pull requests). Thanks

dbmeneses:
Reproduced the problem and tested this fix. Tickets is here: https://jira.sonarsource.com/browse/SONAR-6922. @mvanini I've merged it, thanks!

PR #40, opened 2016-03-17
[JENKINS-33593] Upgrade to new plugin parent pom

armfergom:
Upgrade to new parent pom. Includes many improvements. More info [here](https://github.com/jenkinsci/plugin-pom)

- [x] Upgrade to new parent pom (version 2.3) keeping the same baseline version (1.580.3).
- [x] Fix minor issues found by animal-sniffer plugin.
- [x] Fix minor injected test issues after update to JTH 2.1

Upgrade to new plugin parent pom also integrates FindBug in the build. At the moment it is being ignored (see `properties`in pom.xml) because it finds too many issues to be addressed in this PR.

Apart from some plugin integrations that are useful to increase code quality, the new parent POM is decoupled from the core Jenkins project, both from the Maven and repository perspectives simplifying testing the plugin with different core versions. 

As an example, the Pipeline Plugin (formerly known as workflow) includes this since version [1.14](https://github.com/jenkinsci/workflow-plugin/tree/workflow-1.14).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.580.3</version>
+    <version>2.3</version>
     <relativePath />
   </parent>
 
@@ -101,6 +101,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <hudson.version>${project.parent.version}</hudson.version>
     <sonar.exclusions>target/generated-sources/**/*</sonar.exclusions>
+    <jenkins.version>1.580.3</jenkins.version>
+    <java.level>6</java.level>
   </properties>
 
   <repositories>
@@ -121,8 +123,18 @@
       <!-- needed for SonarPublisher -->
       <groupId>org.jenkins-ci.main</groupId>
       <artifactId>maven-plugin</artifactId>
-      <version>2.7.1</version>
+      <version>2.12.1</version>
       <optional>true</optional>
+      <exclusions>
+        <exclusion>
+	       <groupId>org.apache.httpcomponents</groupId>
+	       <artifactId>httpclient</artifactId>
+        </exclusion>
+        <exclusion>
+           <groupId>org.apache.httpcomponents</groupId>
+           <artifactId>httpcore</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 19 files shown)


src/main/java/hudson/plugins/sonar/client/HttpClient.java
```diff
@@ -37,12 +39,12 @@ public String getHttp(String urlToRead, String username, String password) throws
       if (!StringUtils.isEmpty(password)) {
         userpass = userpass + password;
       }
-      String basicAuth = "Basic " + javax.xml.bind.DatatypeConverter.printBase64Binary(userpass.getBytes(StandardCharsets.UTF_8));
```
dbmeneses: why would you replace Java7 constants with Guava's?

dbmeneses:
Thanks for the contribution. At a first glance, this looks good (apart of a comment). I will test and rebase soon.

src/main/java/hudson/plugins/sonar/client/HttpClient.java
armfergom: With this plugin update, there are some build tools that have been integrated like [animal-sniffer](http://www.mojohaus.org/animal-sniffer/). The baseline being used is 1.580.3 where Java7 is not supported. Because the baseline is 1.580.3, the java level is set to Java6 and animal-sniffer performs the check at that level. [StandardCharsets](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html) do not exist in Java6 so animal-sniffer throw an error and breaks the build.

Possible solutions I can come up with are:

- Bump baseline to 1.642+ and remove `<java.level>6</java.level>`. (Default is Java7)
- Use Google's Charset to avoid this problem.

Chose second option to be less intrusive.

armfergom:
@dbmeneses Thank you for the review!

pom.xml
```diff
@@ -210,30 +204,16 @@
           <artifactId>maven-gpg-plugin</artifactId>
           <version>1.6</version>
         </plugin>
-        <plugin>
-          <groupId>org.jacoco</groupId>
-          <artifactId>jacoco-maven-plugin</artifactId>
-          <version>0.7.5.201505241946</version>
-        </plugin>
       </plugins>
```
dbmeneses: @armfergom the maven plugins' versions should be kept managed. The last version of JaCoCO, for example, is not working properly with SonarQube.

pom.xml
```diff
@@ -259,17 +239,6 @@
...
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <systemPropertyVariables>
-            <!-- We should override default port defined in {@link hudson.UDPBroadcastThread#PORT} to avoid intersections 
-              with real Hudson on Nemo. -->
-            <hudson.udp>33849</hudson.udp>
-          </systemPropertyVariables>
```
dbmeneses: AFAIK, port overlapping could still be a problem when running tests in Jenkins.

dbmeneses:
Thanks for the explanations @armfergom . I've a few more comments. I've relaunched tests in another branch after rebase.

dbmeneses:
I've rebased it, and did minor modifications and merged. Thanks for the contribution.

PR #41, opened 2016-04-20
Allow using of environment variables in project version

MoJo2600:
if you add a "SonarQube Scanner for MSBuild - Begin Analysis" step to your build you have to define a project version. Currently it is not possible to use jenkins environment variables in this field. So you have to define the project version and have to manually edit the build definition if the version changes.
My commit replaces environment variables in the project version. Now you can provide the version number dynamically.

### DIFF
--- a/src/main/java/hudson/plugins/sonar/MsBuildSQRunnerBegin.java
+++ b/src/main/java/hudson/plugins/sonar/MsBuildSQRunnerBegin.java
@@ -120,9 +120,11 @@ public class MsBuildSQRunnerBegin extends AbstractMsBuildSQRunner {
   private void addArgsTo(ArgumentListBuilder args, SonarInstallation sonarInst, EnvVars env, Map<String, String> props) {
     args.add("begin");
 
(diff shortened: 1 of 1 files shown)


henryju:
Thanks for your contribution. @dbmeneses  why not apply the same strategy to projectKey and projectName?

dbmeneses:
Yes, we can apply to all sonar properties. But perhaps we can do it in the SonarBuildWrapper so that we have the same behavior in all builders using the injected variables.
Note that the SonarBuildWrapper is already resolving sonar variables against each other, but not using all the other variables. We can do it there as long as we assume that the other variables are already present when the SonarBuildWrapper is called (this should be true unless other build wrappers or the builders themselves inject more variables). WDYT @MoJo2600 ?

MoJo2600:
I think, that you are right. The SonarBuildWrapper should be the right place, so that the behavior is the same for all builders. I'll try my best to put it there and will update this pull request, ok?

dbmeneses:
Thanks

henryju:
This PR is superseeded by #51. Closing.

PR #42, opened 2016-04-24
[JENKINS-33417] - Fixed circular dependency warning on Startup

Dohbedoh:
@reviewbybees

### DIFF
--- a/src/main/java/hudson/plugins/sonar/SonarGlobalConfiguration.java
+++ b/src/main/java/hudson/plugins/sonar/SonarGlobalConfiguration.java
@@ -49,7 +49,6 @@ public class SonarGlobalConfiguration extends GlobalConfiguration {
 
   public SonarGlobalConfiguration() {
     load();
-    save();
   }
 
   /**


dbmeneses:
How did you came to the conclusion that there is a circular dependency?

Dohbedoh:
This is a warning produced by Jenkins when loading configuration on startup: `Tried proxying hudson.plugins.sonar.SonarGlobalConfiguration to support a circular dependency, but it is not an interface`. (see [JENKINS-33417|https://issues.jenkins-ci.org/browse/JENKINS-33417])

dbmeneses:
Ok thanks

chitraj8:
I see there is latest version sonar-plugin:2.4 will that be having this issue fixed?

Dohbedoh:
@chitraj8 sonar-plugin:2.4 has been released before this fix was merged and therefore does not contain this fix.

dbmeneses:
@chitraj8 v2.4.1 will be released in the next few days and will include the fix.

PR #46, opened 2016-05-26
No public field jdk or getter method throws exception in snipper gene…

eddumelendez:
…rator


janos-ss:
Thank you for your contribution, looking good. I will most probably merge this today.

janos-ss:
A bit more work was needed (forgotten rename in a jelly template), but I could incorporate your change, and now this is in `master`, thanks again!

PR #47, opened 2016-05-26
Replace Computer.getComputer() by workspace.toComputer()

eddumelendez:


kuisathaverat:
that change resolve SONARJNKNS-213 and make Sonar Plugin compatible with Pipeline Plugin

henryju:
This make the plugin technically compatible with the pipeline plugin but there are still many glitches. Especially the name of components/parameters is not properly designed to have a nice DSL. Before officially announcing support of pipeline (and at the same time commiting to keep backward compatibility) I would like to spend some time improving all names.

eddumelendez:
@henryju Should I close this PR since these changes are in your branches? I bet that this is not gonna be merged, right?

henryju:
Sorry guys for not being active on this topic. We are currently busy with other subjects but we'll work on pipeline support as soon as possible. I'll then decide what to do with this P/R.

henryju:
The fix was already applied and FYI we are now working on proper pipeline support. Thanks for your contribution and sorry for the delay.

PR #49, opened 2016-06-17
Fix for MsBuildSQRunner in pipeline

raphaelquati:
Small fix to make MSBuild sonar run in workflow.


janos-ss:
Thank you for bringing this to our attention!

Hi, correct me if I'm wrong, but it seems this PR is more about a workaround for [issue JENKINS-29537 on Jenkins](https://issues.jenkins-ci.org/browse/JENKINS-29537), than necessary logic for the Sonar plugin.

Also, we worked on pipeline support, see the details on this ticket: https://jira.sonarsource.com/browse/SONARJNKNS-254, to be released very soon. If the new release doesn't handle your use case, please explain why, and then we can reopen this PR.

raphaelquati:
Yes. It was a workaround.  I'll watch the ticket.

PR #51, opened 2016-06-23
Better support of environment variables in the MSBuild Scanner config

felipebz:
With this change will be possible to use environment variables for the project key, name and version in the configuration section of "SonarQube Scanner for MSBuild - Begin Analysis".

Please see: https://groups.google.com/d/msg/sonarqube/jW2t_99z5ww/bED6fayoAwAJ


henryju:
Merged, thanks!

N3uRo:
Hi,

In my case I need it also for "Additional arguments" field, can you add it please?

In my workflow I have a project template in Jenkins that I inherit from it so in all my projects I scan them with sonar cube and I need that additional arguments be filled with a variable like "SONAR_EXTRA_ARGS" that I can fill in a child project using this template.

henryju:
@N3uRo  Would be better to open a new P/R so that we don't forget during next sprint.

PR #52, opened 2016-07-01
Update of Confluence help links.

pingunaut:
The names and locations in sonarqube's Confluence changed but the links in this plugin did not which leads to 404 errors if someone clicks the help buttons


henryju:
Manually merged with some modification. Anyway thanks for the contribution.

PR #53, opened 2016-08-25
Extend Sonar icon display name

ochedru:
When a Jenkins job triggers multiple SonarQube analysis (e.g. Maven project with modules), the user cannot distinguish easily between several identical icons. This PR extracts the project name from the URL and adds it to the display name only when there is more than one link.

### DIFF
--- a/src/main/java/hudson/plugins/sonar/action/SonarBuildBadgeAction.java
+++ b/src/main/java/hudson/plugins/sonar/action/SonarBuildBadgeAction.java
@@ -23,6 +23,7 @@ import hudson.plugins.sonar.SonarPlugin;
 
 import hudson.PluginWrapper;
 import hudson.model.BuildBadgeAction;
+import hudson.plugins.sonar.utils.SonarUtils;
 import jenkins.model.Jenkins;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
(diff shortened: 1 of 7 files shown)


henryju:
Hi, can you share you job configuration. I would like to reproduce the issue prior to decide what to do with your P/R. For me a multi-module Maven project should be analyzed once.

ochedru:
Hi, it is difficult to share the job configuration as is, but I can give you a few details:

- We are using maven to build a project with 3 modules written in Java, Javascript and Python (this is a multi-tier application). 
- We have 3 post steps "Execute SonarQube Scanner" configured with 3 different `sonar.properties` files, one per module.
- We have 3 SonarQube projects: one for each module.

Maybe I missed something and I can run a single scanner post step on several modules written in different languages and push the results to a single SonarQube project? Note each module has its own set of unit tests, integration tests and coverage data.

P.S.: I checked out https://github.com/SonarSource/sonar-examples/tree/master/projects/languages/multi-language/multi-language-java-javascript-sonar-runner but this is not a multi-module project.

henryju:
First you should really consider using the SonarQube Scanner for Maven (mvn sonar:sonar). Compared to using the generic SonarQube Scanner (sonar-runner/sonar-scanner) it will populate many properties automatically from your pom.xml.
If you really want to keep using the SonarQube Scanner then I highly suggest you configure a multi-module project to have everything analyzed in one shot. See for example: https://github.com/SonarSource/sonar-examples/tree/master/projects/multi-module/sonar-runner
Running several SQ analysis in the same job is not something we officially support.

ochedru:
OK thanks for the tips! I will give it a try.

PR #67, opened 2016-09-22
Add IT for pipeline with MSBuild

henryju:


its/src/test/java/com/sonar/it/jenkins/orchestrator/JenkinsOrchestrator.java
```diff
@@ -267,14 +270,9 @@ public JenkinsOrchestrator newFreestyleJobWithMaven(String jobName, File project
...
-    
+
     findElement(By.linkText("Invoke top-level Maven targets")).click();
-    try {
-      Thread.sleep(2000);
-    } catch (InterruptedException e1) {
-      e1.printStackTrace();
-    }
-    setTextValue(findElement(By.name("_.targets"), 1), getMavenParams(orchestrator));
+    setTextValue(findElement(driver, By.name("_.targets"), 1), getMavenParams(orchestrator));
```
janos-ss: FYI, the sleep was added as a dirty workaround for a timing issue in ITs executed by cix. Without this sleep, a window will pop-up asking if you want to navigate away from the page with unsaved changes. That said, the sleep was not intended as a solution, just an experiment to see if it indeed helps. It did seem to help, but then we didn't clean it up. If you remove this, probably the build will fail on cix, complaining about a pop-up.

src/main/java/hudson/plugins/sonar/AbstractMsBuildSQRunner.java
```diff
@@ -108,24 +76,23 @@ protected String getExeName(MsBuildSQRunnerInstallation msBuildScanner, EnvVars
...
-  protected static void saveScannerName(Run<?, ?> r, @Nullable String name) throws IOException, InterruptedException {
-    r.addAction(new InjectEnvVarAction(Collections.singletonMap(INST_NAME_KEY, Util.fixNull(name))));
-  }
+  static class SonarQubeScannerMsBuildParams extends InvisibleAction {
 
-  protected static void saveSonarInstanceName(Run<?, ?> r, @Nullable String name) throws IOException, InterruptedException {
-    r.addAction(new InjectEnvVarAction(Collections.singletonMap(SONAR_INST_NAME_KEY, Util.fixNull(name))));
-  }
+    private String scannerName;
+    private String sqServerName;
```
janos-ss: Can't these be `final` ?

henryju: Indeed

its/src/test/java/com/sonar/it/jenkins/orchestrator/JenkinsOrchestrator.java
henryju: ITs are green on my branch on cix so I think the workaround is no more needed

PR #69, opened 2016-10-15
Add @Symbol(sonarScanner) to SonarScanner's Tool Descriptor

eddumelendez:


henryju:
As discussed on the mailing list, we will not make existing freestyle steps compatible with pipeline. Instead we will rely on env inject + sh/bat steps. See https://groups.google.com/forum/#!msg/sonarqube/V6YSZfZ_rH0/a7xpp47TAAAJ. Anyway thanks for your pull request :)

PR #80, opened 2017-04-21
Adding "sonar" Pipeline Step.

gena01:


gena01:
@dbmeneses do you mind reviewing this?

henryju:
Thanks for this contribution but we can't accept it. SimpleBuildStep is not appropriate for long running tasks. See this thread: https://groups.google.com/d/msg/jenkinsci-dev/ehrHDe7SaDE/1AaHzEeqAAAJ
You should use built-in 'sh' or 'bat' steps (we don't want to maintain a specialized copy of those steps).

PR #82, opened 2017-08-10
Fix ITs after SonarJava started to require sonar.java.binaries

henryju:


henryju:
@dbmeneses @janos-ss this is an attempt to fix ITs. Please merge if everything is green

PR #83, opened 2017-09-26
Removed obsolete option

okutane:
`INFO: Option -e/--errors is no longer supported and will be ignored`

We're using latest sonar-scanner for jenkins and I'm seeing the line above in every build log (while we're using default scanner options). I guess we should remove that option.


okutane:
@janos-ss can you review this please?

janosgyerik:
I'm on holiday, but maybe @henryju or @dbmeneses can ;-)

henryju:
Thanks for your contribution. Ticket: https://jira.sonarsource.com/browse/SONARJNKNS-288

PR #85, opened 2017-10-19
Do not parse build log but explicitly tell the Maven plugin where to put its report

brenuart:
This PR is all about avoiding to parse the build logs to discover where the Maven plugin may have put its report file. In short the modified version does the following:
- Ask Jenkins for a temporary directory alongside the workspace to host the report generated by the Maven sonar plugin.
- Use `sonar.working.directory` to tell the Maven plugin where to put its report file.
- Consume the report file during `tearDown()` from that location without having to parse the build log to find it back...
- Delete the report folder on completion

### DIFF
--- a/src/main/java/hudson/plugins/sonar/SonarBuildWrapper.java
+++ b/src/main/java/hudson/plugins/sonar/SonarBuildWrapper.java
@@ -144,6 +144,9 @@ public class SonarBuildWrapper extends SimpleBuildWrapper {
 
     map.put("SONAR_EXTRA_PROPS", getOrDefault(getAdditionalProps(inst), ""));
 
+    String workingDir = reportDir.getRemote();
+    map.put("SONAR_WORKING_DIRECTORY", workingDir);
+    
     StringBuilder sb = new StringBuilder();
     sb.append("{ \"sonar.host.url\" : \"").append(StringEscapeUtils.escapeJson(hostUrl)).append("\"");
     if (!login.isEmpty() || !token.isEmpty()) {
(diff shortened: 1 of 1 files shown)


henryju:
I have chosen a different approach, not trying to change the sonar.working.directory location. I'm basically doing a full search in the workspace for **/report-task.txt. This is consistent with what we have done in the VSTS extension. Anyway thanks for raising the issue, and proposing a fix.

Dabassman:
With this solution, I am facing the problem that the sonar scanner is always finding more than one report-task.txt.
We have 5 Jenkins builds with sonar plugin which are build one after another in the same workspace (having their own subdirectory).
Since the plugin starts in the workspace folder, it finds all five files. Can I somehow tell the plugin which txt file to use?

Dabassman:
Standard project; pulling sources from SVN, MSBuild for c#

brenuart:
So your project is a multi-module maven project and the build is launched from the top level pom?

Dabassman:
I have five build elements in jenkins but only one build processor.
Every build checks out to %workspace%\Project[1..5]

After the job for Project1, the job for Project2 starts.

The SQ plugin searches for the task file in %workspace%.

brenuart:
Oh... so you have 5 different jobs sharing the same workspace?
If that’s the case, I don’t think the plugin is designed to support this scenario...

Dabassman:
But how should I split that? The workspace is defined per build processor; if you have more projects than build processors, projects will share the workspace.
Maybe I am misunderstanding something...

brenuart:
This would not be a problem if the plugin was using a temp dir outside of the workspace (unique temp folder assigned and cleaned by Jenkins at the start of the build). This does not seem to be the case.

In the meantime i would suggest you clean the entire workspace at the start of each build. This way you would not be affected by files left over by previous builds.

You can ask your git/svn plugin to perform a so-called clean checkout. You can also install and use another plugin (i forgot its name) who’s purpose is to clean/delete the workspace content.

Dabassman:
Thank you for your help. I have found an option to change the workspace directory per job (extended settings); this seems to do the trick.

brenuart:
Usually each job gets its own workspace... by default.

PR #86, opened 2018-02-02
Add Chinese translation

LinuxSuRen:


LinuxSuRen:
@henryju Can you help me to review this PR, thank you.

henryju:
I'm sorry but I have to reject this pull request. First it is translating only a very small subset of the plugin so I don't see the point. And more importantly we tend to avoid translating our products since we are not sure we can maintain that. There is currently no native Chinese speaker at SonarSource :) On SonarQube side this is a bit different, since i18n come from third party plugins, that can have their own lifecycle, and are maintained fully by the community.

LinuxSuRen:
I just think of it will be helpful to millions of Chinese customer, if I do this. But it's your call finally.

henryju:
Either people can read English, and we don't need translation, or they don't, and we need to target near 100% translation. I fail to see the benefit of translating a small part only.

LinuxSuRen:
I will translate all of them if you can accept.

henryju:
Fine, then I'm reopening the PR. Please complete it. FYI we plan to do a release next week.

Evangelink:
Hi @LinuxSuRen,

Out of curiosity could you manage to provide the full translation PR soon as we will have a new release of the scanner for jenkins?

LinuxSuRen:
@Evangelink Sorry about this, I'm little busy recently. I should close this PR for now, and I will reopen after finished.

PR #87, opened 2018-02-22
Changed 'locahost' to 'localhost' in default_server_url

chrrie:
When configuring the jenkins plugin it states "Default is http://localhost:9000", but if the Server URL field is left blank 'locahost:9000' is used - causing the build pipeline to fail with an exception - unknown host - and a stack trace. 

I have version 2.6.1 installed, Jenkins 2.108.


henryju:
Good catch, thanks!

PR #92, opened 2018-04-20
added trim to getServerAuthenticationToken

Sravan-yarlagadda:
added trim to getServerAuthenticationToken to remove the leading and trailing spaces

### DIFF
--- a/src/main/java/hudson/plugins/sonar/SonarInstallation.java
+++ b/src/main/java/hudson/plugins/sonar/SonarInstallation.java
@@ -138,10 +138,7 @@ public class SonarInstallation implements Serializable {
    * @since 2.4
    */
   public String getServerAuthenticationToken() {
-    if (serverAuthenticationToken != null) {
-      return serverAuthenticationToken.trim();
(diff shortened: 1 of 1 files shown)


Sravan-yarlagadda:
Hi @henryju Can you please review the changes.

Sravan-yarlagadda:
@henryju  I have updated the change as per you comments. Added trim in the get method and updated the Junit test method to test the current change.

src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -138,6 +138,9 @@ public String getServerUrl() {
    * @since 2.4
    */
   public String getServerAuthenticationToken() {
+    if (serverAuthenticationToken != null) {
+      return serverAuthenticationToken.trim();
```
henryju: Please simplify this method using `return StringUtils.trimToNull(serverAuthenticationToken);`

henryju:
Merged, thanks

PR #94, opened 2018-06-01
Replace SonarQube by Sonar in the different names/texts

Evangelink:


Evangelink:
@henryju I have renamed only the Scanner for MSBuild part as it is the only one for which the job was done everywhere + the only one I am really sure how to test.

I would still be happy if you could make one update test (I tried locally and it was successful) to be sure I didn't broke anything.

PR #101, opened 2018-07-10
Fix/sonarjnkns 256: Support credentials plugin

luka5:
This fixes the [SONARJNKNS-256](https://jira.sonarsource.com/browse/SONARJNKNS-256). We use the Jenkins credentials plugin instead of saving the token directly inside the xml file. The tests are extended, updated and running. A manually created `hpi` file works as expected.

The changes are not compatible to the previous version, since the former `serverAuthenticationToken` gets removed completely. Is there a way to provide a migration strategy? We think, creating credentials automatically is not a proper way to handle that.

image

`hudson.plugins.sonar.SonarGlobalConfiguration.xml`:
```xml
<?xml version='1.0' encoding='UTF-8'?>
<hudson.plugins.sonar.SonarGlobalConfiguration plugin="sonar@2.X">
  <installations>
    <hudson.plugins.sonar.SonarInstallation>
      <name>SonarQube Server</name>
      <serverUrl></serverUrl>
      <credentialsId>sonarqube_token</credentialsId>
....
```

What's your opinion on these changes? Do you want us to change something? We are already using it and it works exactly the way we want.

### DIFF
--- a/src/main/java/hudson/plugins/sonar/SonarInstallation.java
+++ b/src/main/java/hudson/plugins/sonar/SonarInstallation.java
@@ -153,6 +153,12 @@ public class SonarInstallation implements Serializable {
   }
 
   /**
+   * @since X
+   */
+  @SuppressWarnings("unused")
+  public String getCredentialsId() { return credentialsId; }
+
+  /**
    * @return version of sonar-maven-plugin to use
    * @since 1.5
    */
--- a/src/main/java/hudson/plugins/sonar/SonarInstallation.java
+++ b/src/main/java/hudson/plugins/sonar/SonarInstallation.java
@@ -19,7 +19,11 @@
  */
 package hudson.plugins.sonar;
 
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.common.PasswordCredentials;
+import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import hudson.AbortException;
+import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.plugins.sonar.model.TriggersConfig;
 import java.io.Serializable;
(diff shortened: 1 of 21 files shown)


src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -37,9 +41,9 @@
   private final String serverUrl;
 
   /**
-   * @since 2.4
+   * @since X
```
henryju: To be fixed

src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -135,12 +139,25 @@ public String getServerUrl() {
...
+    if (cred == null) { return null; }
+    return ((PasswordCredentials) cred).getPassword().getPlainText();
   }
 
+  public StandardCredentials getCredentials(Run<?, ?> build) {
+    return CredentialsProvider.findCredentialById(credentialsId, StandardCredentials.class, build);
+  }
+
+  /**
+   * @since X
```
henryju: TBD

henryju:
@drautureau-sonarsource  What you you think of the proposed feature?

src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -135,12 +139,25 @@ public String getServerUrl() {
   }
 
   /**
-   * @since 2.4
+   * @since X
```
luka5: TBD

luka5:
@henryju Thanks for the feedback. Yes, I totally forgot about the integration tests. Could you please give me some hints on how I can execute these?

I also asked for hints about migrating the token as a credential in the [dev mailing list.](https://groups.google.com/forum/#!topic/jenkinsci-dev/gv8gyJ4XKmk)
I'll try to follow a similar approach as developed in [subversion-plugin](https://github.com/jenkinsci/subversion-plugin)

henryju:
To run ITs, basic command is:
```
cd its
mvn -Djenkins.runtimeVersion=2.89.4 -Dsonar.runtimeVersion=7.2 clean verify
```
If you have troubles, let me know.

Also, based on Stephen feedback, I wonder if we should not introduce our own credential type (SonarToken ?)

luka5:
closed in favor of #101

PR #107, opened 2018-09-21
Make Jenkins call the SQ URL defined in the SonarInstallation

thomasgl-orange:
Make Jenkins call the SQ URL defined in the SonarInstallation, instead of SQ public URL.

see https://community.sonarsource.com/t/sonarjnks-2-8-jenkins-plugin-calls-public-sonarqube-url-instead-of-the-one-set-in-sonarinstallation/2776?u=thomasgl-orange

To make this work with the corner-case requirements from SONARJNKNS-287 (server URL defined via some variables), we store the expanded URL in SonarAnalysisAction.

Note that vars expansion for the URL is not implemented in SonarPublisher/SonarMaven (would be possible, but it's deprecated code, so we keep its old behavior unchanged).

### DIFF
--- a/its/src/test/java/com/sonar/it/jenkins/JenkinsTest.java
+++ b/its/src/test/java/com/sonar/it/jenkins/JenkinsTest.java
@@ -38,6 +38,9 @@ import static org.fest.assertions.Assertions.assertThat;
 
 public class JenkinsTest {
 
+  // Jenkins should only talk to the URL actually configured in a SonarInstallation
+  final static String SONAR_PUBLIC_BASE_URL = "http://sonar_public_domain/";
+
   @ClassRule
   public static Orchestrator orchestrator = JenkinsTestSuite.ORCHESTRATOR;
 
@@ -47,7 +50,7 @@ public class JenkinsTest {
   @BeforeClass
   public static void setUpSonar() {
     // Workaround for SONAR-4257
-    orchestrator.getServer().getAdminWsClient().update(new PropertyUpdateQuery("sonar.core.serverBaseURL", orchestrator.getServer().getUrl()));
+    orchestrator.getServer().getAdminWsClient().update(new PropertyUpdateQuery("sonar.core.serverBaseURL", SONAR_PUBLIC_BASE_URL));
   }
 
   @BeforeClass
@@ -87,6 +90,7 @@ public class JenkinsTest {
     waitForComputationOnSQServer();
     assertThat(getProject(projectKey)).isNotNull();
     assertSonarUrlOnJob(jobName, projectKey);
+    jenkins.assertQGOnProjectPage(jobName);
   }
 
   @Test
@@ -100,6 +104,7 @@ public class JenkinsTest {
     waitForComputationOnSQServer();
     assertThat(getProject(projectKey)).isNotNull();
     assertSonarUrlOnJob(jobName, projectKey);
+    jenkins.assertQGOnProjectPage(jobName);
   }
 
   @Test
@@ -188,12 +193,12 @@ public class JenkinsTest {
   }
 
   private void assertSonarUrlOnJob(String jobName, String projectKey) {
-    assertThat(jenkins.getSonarUrlOnJob(jobName)).startsWith(orchestrator.getServer().getUrl());
     if (jenkins.getServer().getVersion().isGreaterThanOrEquals(2, 0)) {
       assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(URLEncoder.encode(projectKey));
     } else {
       assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(projectKey);
     }
+    assertThat(jenkins.getSonarUrlOnJob(jobName)).startsWith(SONAR_PUBLIC_BASE_URL);
   }
 
   private static void waitForComputationOnSQServer() {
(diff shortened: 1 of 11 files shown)


thomasgl-orange:
Note that I've been having integration tests errors which were not related to my own changes, in `com.sonar.it.jenkins.JenkinsTest`.  So I've actually been testing this with one additional change, not included here:

```
--- a/its/src/test/java/com/sonar/it/jenkins/JenkinsTest.java
+++ b/its/src/test/java/com/sonar/it/jenkins/JenkinsTest.java
@@ -193,12 +193,15 @@ public class JenkinsTest {
   }
 
   private void assertSonarUrlOnJob(String jobName, String projectKey) {
-    if (jenkins.getServer().getVersion().isGreaterThanOrEquals(2, 0)) {
-      assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(URLEncoder.encode(projectKey));
-    } else {
-      assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(projectKey);
-    }
     assertThat(jenkins.getSonarUrlOnJob(jobName)).startsWith(SONAR_PUBLIC_BASE_URL);
+    // FIXME: I don't get an URL-encoded suffix (Jenkins 2.121.3, SonarQube 6.7.5)
+    // And also JenkinsWithoutMaven makes a not-URL-encoded assertion.
+    //if (jenkins.getServer().getVersion().isGreaterThanOrEquals(2, 0)) {
+    //  assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(URLEncoder.encode(projectKey));
+    //} else {
+    //  assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(projectKey);
+    //}
+    assertThat(jenkins.getSonarUrlOnJob(jobName)).endsWith(projectKey);
   }
 
   private static void waitForComputationOnSQServer() {
```

thomasgl-orange:
Another note... To quickly "visualize" what I'm trying to fix here, you can revert these two chunks, and then run integration tests:

```
--- a/src/main/java/hudson/plugins/sonar/action/SonarCacheAction.java
+++ b/src/main/java/hudson/plugins/sonar/action/SonarCacheAction.java
@@ -73,7 +73,7 @@ public class SonarCacheAction extends InvisibleAction {
       return cached;
     }
 
-    ProjectInformation proj = resolver.resolve(analysis.getServerUrl(), analysis.getUrl(), taskId, analysis.getInstallationName());
+    ProjectInformation proj = resolver.resolve(analysis.getUsableServerUrl(), analysis.getUrl(), taskId, analysis.getInstallationName());
     if (proj != null) {
       infoByTaskId.put(taskId, proj);
     }
```

```
--- a/src/main/java/org/sonarsource/scanner/jenkins/pipeline/WaitForQualityGateStep.java
+++ b/src/main/java/org/sonarsource/scanner/jenkins/pipeline/WaitForQualityGateStep.java
@@ -150,7 +150,7 @@ public class WaitForQualityGateStep extends Step implements Serializable {
       for (SonarAnalysisAction a : reversedActions) {
         ceTaskId = a.getCeTaskId();
         if (ceTaskId != null) {
-          serverUrl = a.getServerUrl();
+          serverUrl = a.getUsableServerUrl();
           installationName = a.getInstallationName();
           break;
         }
```

Reverting the first one will break the Quality Gate report on the Jenkins jobs pages (and you will see failed HTTP requests to sonar_public_domain in the Jenkins master log).

Reverting the second one will break the `waitForQualityGate` Pipeline tests (again, because of failed HTTP requests to some random domain).

henryju:
Hi,

Thanks for reporting the issue. I think this is a consequence of https://jira.sonarsource.com/browse/SONARJNKNS-287. I gave only a quick look but I think your proposed change will reintroduce the issue of SONARJNKNS-287. So we have to find a better way to handle this.

thomasgl-orange:
Hi Julien,

I don't understand why you're closing this PR. Can you please reconsider?

If that's the reason, sorry I did not reply to your previous comment, here is what I should have said:

> I think this is a consequence of jira.sonarsource.com/browse/SONARJNKNS-287.

Yes, exactly what I wrote in the PR text.

> I think your proposed change will reintroduce the issue of SONARJNKNS-287.

You had introduced a test for that in 28cbd920: `SonarBuildWrapperTest.testEnvironment()`. It works as expected with this PR. Isn't that enough?

> So we have to find a better way to handle this.

I'm open to discussing it, but so far, I think this PR improves the situation, without .

Your fix to SONARJNKNS-287 was: "_Let's use the `SonarInstallation` URL to begin with, substituting vars if any, and then later continue the Jenkins->SQ interactions with a completely different URL which we happen to have at hand because the one from `SonarInstallation` might actually include vars which we are not able to substitute anymore._"

My fix is (at least in intents / unless I've overlooked something): "_Let's use the `SonarInstallation` URL to begin with, substituting vars if any, and then later continue the Jenkins->SQ interactions with the exact same URL (which will have been saved in expanded form in the `SonarAnalysisAction`, to avoid ever having to substitute vars again)._"

To me, it sounds like a reasonable way to deal with the weird corner-case scenario from SONARJNKNS-287, while still honouring whatever has been explicitly configured as the SQ URL in the `SonarInstallation`. And as you can guess, I'm on an environment where Jenkins instances don't have access to SQ public URLs (OK, that's a weird corner-case too), so the 2.8.1 plugin without this patch was a major regression for us.

Please let me know if you want me to merge master in the PR, or if there's anything else I could do which could help moving forward.

Regards,
Thomas.

thomasgl-orange:
Thanks Henry for reopening.

Because there were many conflicts to fix in this PR for updating it for 2.9, and I didn't know how you prefer to get PR updates (additional merge commits, or rebase+force), I've submitted an updated patch in #123; feel free to close this one again and keep #123, or to ask me to update this one and close the other.

PR #110, opened 2019-02-13
SONARJNKNS-256 Support credentials plugin

timja:
Replaces https://github.com/SonarSource/sonar-scanner-jenkins/pull/101

I took https://github.com/SonarSource/sonar-scanner-jenkins/pull/101 as a base, resolved the conflicts and added credential migration based on the slack-plugin migration

The change has been manually tested and well tested with unit tests


src/test/java/hudson/plugins/sonar/BaseTest.java
```diff
@@ -108,17 +110,20 @@ protected void setBuildResult(Project<?, ?> project, Result result) throws Excep
   public void testPassword() throws Exception {
     configureDefaultSonarRunner(false);
     configureSecuredSonar();
+
     FreeStyleProject project = setupFreeStyleProjectWithSonarRunner();
     Run<?, ?> build = build(project);
 
     assertLogContains("sonar-runner", build);
-    assertLogDoesntContains("sonar.login", build);
+    assertLogContains("-Dsonar.host.url=http://example.org:9999/sonar ********", build); // check masked token is in logs
```
timja: This test previously passed when the token wasn't even being added (I wasn't adding the credential properly)
This change asserts that some masked thing goes in, in the expected place

src/main/resources/hudson/plugins/sonar/SonarGlobalConfiguration/config.jelly
```diff
@@ -20,9 +18,9 @@
           <f:entry title="${%ServerUrl}" description="${%ServerUrlDescr}">
             <f:textbox name="sonar.serverUrl" value="${inst.getServerUrl()}"/>
           </f:entry>
-                    
-          <f:entry title="${%ServerToken}" description="${%ServerTokenDescr}">
-            <f:password name="sonar.serverAuthenticationToken" value="${inst.getServerAuthenticationToken()}"/>
+
+          <f:entry title="${%ServerToken}" field="credentialsId">
```
henryju: Is it required to remove the description attribute? And if yes, don't forget to clean resource bundles.

timja: No not at all, I didn't actually touch this file, it was from the previous PR.
I've restored it now

Screenshot of what it looks like:
image

src/test/java/hudson/plugins/sonar/BaseTest.java
timja: I've added an extra assertion looking for the secret to be extra sure it doesn't go in the log

timja:
Looking into the test failures, both of `GlobalCredentialMigratorTest` pass locally for me

henryju:
Any chance you could rebase on master and fix conflicts?

timja:
This branch is up to date with master, can you rebase on merge please? there would be quite a lot of conflicts to resolve

timja:
All tests are passing except the new credential one,
fails when all tests are run, but passed when run individually

Fails:
```
mvn clean test
```

Passes
```
mvn clean test -Dtest=hudson.plugins.sonar.GlobalCredentialMigratorTest -Dmaven.javadoc.skip
```

Also passes in IntelliJ

Looking into it

timja:
@henryju I've removed the failing test, I've tried quite a few different approaches and I can't see any reason why the secret is failing to decrypt when run with more than one test

Any thoughts?

timja:
Test is back in, I found a fix by searching github:
https://github.com/jenkinsci/nexus-platform-plugin/blob/f6b4592749034967f7f49ae75e04c3d0bccd6dd5/src/test/resources/org/sonatype/nexus/ci/config/ComToOrgMigratorIntegrationTest/credentials.xml

Apparently the xml will take a plain text value and then encrypt it for you.

That worked fine both individually and when all are run

timja:
I've rebased all my commits but @luka5 's ones alone

timja:
@henryju Anything left on this?

henryju:
Time :)

More seriously, we have discussed this again today, and I will try to plan a Jenkins sprint in the coming month.

luka5:
Thanks for your support!
1. Would you like me to sign my commits?
2. May I close https://github.com/SonarSource/sonar-scanner-jenkins/pull/101 ?

timja:
@luka5 
1. up to you, I would say no need
2. yes 👍

luka5:
> @luka5
> 
> 1. up to you, I would say no need
> 2. yes +1

ok, alright. I'll keep it like this.

henryju:
Replaced by #112

PR #120, opened 2019-06-24
build with java11

xavierbourguignon:


pom.xml
```diff
@@ -50,7 +50,7 @@
     <sonar.exclusions>target/generated-sources/**/*</sonar.exclusions>
     <gitRepositoryName>sonar-scanner-jenkins</gitRepositoryName>
     <version.artifactory.plugin>2.6.1</version.artifactory.plugin>
-    <jenkins.version>2.89.4</jenkins.version>
+    <jenkins.version>2.176.1</jenkins.version>
```
henryju: Is this change really required? It will be the minimal version of Jenkins supported by the plugin (at least in the update center). By experience we know people are not fast to upgrade their Jenkins, so the older version we support, the better.

xavierbourguignon: oldest supported version is 2.150.3
with older version there is a compilation failure:
> java.nio.file.NoSuchFileException: sonar-scanner-jenkins/target/classes/META-INF/annotations/hudson.Extension

henryju: OK, this one was released on 2019-02-27, that's probably more acceptable.

henryju:
Merged, thanks

PR #122, opened 2019-07-26
Fix formatting

jblaine:


henryju:
@ganncamp can I let you review this (looks harmless to me but you are in a better position to know) ;)

henryju:
Merged, thanks

PR #123, opened 2019-08-12
Use SQ URL from SonarInstallation, not public URL

thomasgl-orange:
Make Jenkins call the SQ URL defined in the `SonarInstallation`, instead of SQ public URL. To make this work with the corner-case requirements from [SONARJNKNS-287](https://jira.sonarsource.com/browse/SONARJNKNS-287) (`SonarInstallation` URL defined via some variables), we store the var-expanded URL in `SonarAnalysisAction`, and reuse this URL in `waitForQualityGate`.  If the `SonarAnalysisAction` has been stored without a ready-to-use server URL (for instance if it contains variables we could not expand, then the SQ public URL (from `report-task.txt`) will still be used as a fallback.

This PR is an updated version (for v2.9 - and now for v2.10) of something I had proposed already for v2.8.x in #107.


pierre-guillot-sonarsource:
hello @henryju :) i know you are not responsible of this anymore, but i could use your help here 👍any feedback on this ?

henryju:
Needs to verify all ITs are still passing, and also do some manual testing I guess ;)

pierre-guillot-sonarsource:
I'm closing this as the PR looks no more active. Please comment/reopen if needed :)

thomasgl-orange:
> I'm closing this as the PR looks no more active. 

Did you expect more activity on my side to consider this PR? If so, please let me know what I could do.

Just to give you some context, the bug it fixes would be a blocker where I work, because we can't access the SQ public URL from our Jenkins servers. Which has been a perfectly acceptable situation for years (we simply had to set the internal URLs in the SonarInstallation configurations of our Jenkins servers), until it became an issue with v2.8 because of changes made for SONARJNKNS-287.  We are thus using a patched version of the SQ plugin (we've first been using a patched 2.8.x with #107, and we are now using 2.9 with #123). So, this patch, or its previous iteration, are actually in use already on hundreds of Jenkins servers, and have been since September 2018. Of course, I would love to see it merged in a future release of the plugin.

pierre-guillot-sonarsource:
@dbmeneses can you please have a look ? :)

thomasgl-orange:
Updated the PR for v2.10. There was only one simple merge conflict, in `SonarPluginTest#setUp()`.

I've successfully re-run the `its` tests with:
- OpenJDK 11.0.4 (Linux)
- `export JENKINS_VERSION=2.190.2`
- `-Dsonar.runtimeVersion=7.9.1`

thomasgl-orange:
Now also merged `master` (no conflict), to get eae7db7 and be able to run IT tests on 6.7.x again.  Tests were successful again, this time with this combination:
- OpenJDK 1.8.0_222 (Linux)
- `export JENKINS_VERSION=2.164.3`
- `-Dsonar.runtimeVersion=6.7.7`

usmonster:
Hello! I'd like to bump this, as it's still a blocker for my organization, and we'd rather not have to patch it ourselves.

@thomasgl-orange, do you think you can do a quick rebase and push just to trigger the CI again so it runs to completion?

@pierre-guillot-sonarsource, @dbmeneses, can you please review it and approve if it looks good?

Thanks!

pierre-guillot-sonarsource:
@mark-rekveld-sonarsource Hey Mark, could you review this, please? This external PR has been open for way too long here :)

thomasgl-orange:
I've merged master again (no conflict). FYI, we're still using this patch in my org, currently on top of 2.11. Still working fine our use cases.

I'm currently running integration tests. So far, I have one failure, on `SonarPluginTest.qualitygate_pipeline_failed_with_unknown_webhook_secret_id` (`waitForQualityGate` succeeds whereas it is expected to fail because of bad secret id), and I suspect it is indeed a regression with this patch, so I'm looking into it.

src/main/java/hudson/plugins/sonar/action/SonarAnalysisAction.java
```diff
@@ -117,4 +146,11 @@ public String getInstallationName() {
   public String getCredentialsId() {
     return credentialsId;
   }
+
+  @CheckForNull
+  @Transient
+  public String getUsableServerUrl() {
```
usmonster, line 152: I might name it `getServerDisplayUrl()` or maybe just `getDisplayUrl()`?

thomasgl-orange: > I might name it getServerDisplayUrl() or maybe just getDisplayUrl()?

It's the opposite: what I've called the "usable" server URL is the URL Jenkins should actually "use" (the right URL for calling SQ webservices from Jenkins), which might be different from the public SQ URL (the one you would display to the end-user). It's the one you would configure in your `SonarInstallation` ("SonarQube servers" in the Jenkins global configuration).

But yes, I'm not super happy with this name either. I didn't want to call it "internal URL", because on m setup it's just the same as the public URL, so it's not always "internal". Should really be `getTheRightBaseURLForCallingSQWebservices()` :)
I'm open to better suggestions, of course.

henryju: What about `getServerApiEndpoint()`?

usmonster: Ah, it's clear now. In that case, I'd call it `getServiceUrl()`, though maybe too close to "ServerUrl"?

thomasgl-orange: Thanks @henryju, it's now `getServerApiEndpoint()`: 3b28a81

thomasgl-orange:
> So far, I have one failure, on `SonarPluginTest.qualitygate_pipeline_failed_with_unknown_webhook_secret_id` (`waitForQualityGate` succeeds whereas it is expected to fail because of bad secret id)

[This test](https://github.com/SonarSource/sonar-scanner-jenkins/blob/master/its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java#L429) is actually flaky, it's not a regression with this PR.

See [`WaitForQualityGateStep#start()`](https://github.com/SonarSource/sonar-scanner-jenkins/blob/master/src/main/java/org/sonarsource/scanner/jenkins/pipeline/WaitForQualityGateStep.java#L166): 
- in `processStepParameters()`, the validity of the `webhookSecretId` is not tested, so no failure at this point
- then comes `checkTaskCompleted()`, with two nominal outcomes:
  1. `false`: the task is pending on SQ. Then `waitForQualityGate` pauses, waiting for a callback on the `sonarqube-webhook` URL.
  2. `true`: the task has already been processed successfully on SQ. Then `waitForQualityGate` finishes in success.
- in the former case, the call from SQ to the webhook will lead to executing [`onComplete(...)`](https://github.com/SonarSource/sonar-scanner-jenkins/blob/master/src/main/java/org/sonarsource/scanner/jenkins/pipeline/WaitForQualityGateStep.java#L267), and thus [`validateWebhook(...)`](https://github.com/SonarSource/sonar-scanner-jenkins/blob/master/src/main/java/org/sonarsource/scanner/jenkins/pipeline/WaitForQualityGateStep.java#L299), which is where the `webhookSecretId` is finally verified. In this test (with `unknownSecret`), it will fail because `A webhook secret id was configured, but the corresponding credential could not be found`, which is the expected outcome.
- in the later case, the Pipeline will end immediately in success, which is not expected by this test.

So, basically, this test only works when SQ is slow enough to process the task (or, from a different pov, Jenkins is fast enough to do its initial check of the task status before it has been processed by SQ).

I don't know if the right fix is to modify `waitForQualityGate` to check the `webhookSecretId` earlier (even if it might actually never be used), or to somehow modify the test case to ensure things happen in the right order (I'm not sure how).  Anyway, that's not something to be dealt with in this PR.


Other than that, integration tests seems actually fine with this PR. It's been a bit painful because of sporadic errors and timeouts due to slow/busy computer/harddrive/internet, but in the end I've successfully tested several setups:
- OpenJDK 8, SQ 6.7.7, Jenkins 2.222.4
- OpenJDK 11, SQ 7.9.3, Jenkins 2.190.3
- OpenJDK 11, SQ 8.3.1.34397, Jenkins 2.222.4

twolfart:
This is hindering me, too. Resolving this would be helpful for environments depending on different internal and external urls.

@thomasgl-orange Thank you for providing and maintaining the PR over this long time.

thomasgl-orange:
> This is hindering me, too. Resolving this would be helpful for environments depending on different internal and external urls.
> 
> @thomasgl-orange Thank you for providing and maintaining the PR over this long time.

@twolfart  Well, I've got some good news for you, apparently version 2.12 of the sonar plugin includes some changes by @mark-rekveld-sonarsource similar to what I had proposed in this PR: 8c23330b7408a9b33750d25149b97932afc5215d

From a quick look at the diff:
- the idea is exactly the same (storing a var-expanded version of the requested SQ URL in the `SonarAnalysisAction`),
- some additional minor refactoring was done (simplified `SonarUtils.addBuildInfoTo(...)` signature)
- var expansion of the URL seems slightly wrong in the case of Pipeline jobs (but really, this var-expansion thing from [SONARJNKNS-287](https://jira.sonarsource.com/browse/SONARJNKNS-287), which introduced the bug in the first place, was such a weird requirement, I wouldn't really care)
- there is no integration tests (in this PR, I was forcing a broken public URL on the SQ server, to check across all IT cases that it was actually never used by Jenkins)

I think the best result could have been achieved simply by posting a review here instead of reauthoring the whole fix, and I must admit I'm a little bitter about how this external contribution was dealt with by SonarSource overall, but anyway, let's see the positive side of things, two years after #107, this annoying bug is finally fixed.

pierre-guillot-sonarsource:
> and I must admit I'm a little bitter about how this external contribution was dealt with by SonarSource overall

That's a fair point, and I'm personally not happy either. Such requests are currently "hard" to follow-up because before @mark-rekveld-sonarsource joined we had no DRI on scanners for more than a year at SonarSource. So the situation should improve, now that we have people on it.

Anyway, again, thanks for your contribution :)

PR #124, opened 2019-08-26
Fix serverAuthenticationToken being required for configuration via CasC

Smasherr:
This fix avoids the following error message

```io.jenkins.plugins.casc.ConfiguratorException: serverAuthenticationToken is required to configure class hudson.plugins.sonar.SonarInstallation```

when using a YAML configuration without `serverAuthenticationToken`

### CODE REVIEW

File: src/test/java/hudson/plugins/sonar/CasCTest.java [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+    @Test
+    public void testMissingServerAuthenticationToken() throws Exception {
+        URL config = this.getClass().getResource("CasCTest/config.yaml");
+        ConfigurationAsCode.get().configure(config.toExternalForm());
+
+        SonarInstallation installation = GlobalConfiguration.all()
+                .get(SonarGlobalConfiguration.class)
+                .getInstallations()[0];
+        Assert.assertEquals("test-id", installation.getCredentialsId());
```
henryju: We are using assertj for better readability: `assertThat(installation.getCredentialsId()).isEqualTo("test-id");`

oleg-nenashev:
CC https://github.com/jenkinsci/configuration-as-code-plugin/issues/809

src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -76,24 +77,23 @@ public SonarInstallation(String name,
...
       mojoVersion, additionalProperties, additionalAnalysisProperties, triggers);
+    this.setServerAuthenticationToken(Secret.fromString(StringUtils.trimToNull(serverAuthenticationToken)));
   }
 
   @DataBoundConstructor
   public SonarInstallation(
     String name,
     String serverUrl,
     @Nullable String credentialsId,
-    @Nullable Secret serverAuthenticationToken,
```
oleg-nenashev: Breaks binary compatibility

src/test/java/hudson/plugins/sonar/CasCTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Test;
+
+import java.net.URL;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CasCTest extends SonarTestCase {
+
+    @Test
+    public void testMissingServerAuthenticationToken() throws Exception {
```
oleg-nenashev: See https://github.com/jenkinsci/antisamy-markup-formatter-plugin/pull/13 for a better test pattern

src/main/java/hudson/plugins/sonar/SonarInstallation.java
oleg-nenashev: Afaik it should work OOTB if you switch to `@CheckForNull`

Smasherr: I tried to revert my constructor change and change from `@Nullable` to `@CheckForNull`, but `CasCTest` failed.

```io.jenkins.plugins.casc.ConfiguratorException: serverAuthenticationToken is required to configure class hudson.plugins.sonar.SonarInstallation```

Comment (oleg-nenashev) [reply src/main/java/hudson/plugins/sonar/SonarInstallation.java]: https://github.com/jenkinsci/configuration-as-code-plugin/blob/f72058445c0250041303a73b5d8566140d337c67/plugin/src/main/java/io/jenkins/plugins/casc/impl/configurators/DataBoundConfigurator.java#L109-L137 BTW. It is actually an inverse way.

I;d guess the rejection is caused by the package-level anotation: https://github.com/SonarSource/sonar-scanner-jenkins/blob/22fd2881d4a8e247cc50fb79f80c67c2865f4d90/src/main/java/hudson/plugins/sonar/package-info.java which is enforced even when `CheckForNull` is set. I will create a ticket for JCasC

Comment (oleg-nenashev) [reply src/main/java/hudson/plugins/sonar/SonarInstallation.java]: https://github.com/jenkinsci/configuration-as-code-plugin/issues/1025 . Better to fix in JCasC IMHO

Comment (Smasherr) [reply src/main/java/hudson/plugins/sonar/SonarInstallation.java]: If I comment out `@ParametersAreNonnullByDefault` from `package-info.java` the test passes with both annotations `@Nullable` and `@CheckForNull`. So I guess that actually should be resolved in CasC?

Comment (oleg-nenashev) [reply src/main/java/hudson/plugins/sonar/SonarInstallation.java]: Yup. Not a big deal for this particular issue

Comment (Smasherr) [reply src/main/java/hudson/plugins/sonar/SonarInstallation.java]: I will close this PR then. Well, at least I learned about the correct test pattern ;-)

**Comment by Smasherr** [comment]
Should be fixed in CasC https://github.com/jenkinsci/configuration-as-code-plugin/issues/1025

**Comment by thorin** [comment]
jenkinsci/configuration-as-code-plugin#1025 has been merged.

But now `@Nullable` should be replaced with `@CheckForNull` on SonarInstallation to fix the configuration via CasC.

**Comment by Smasherr** [comment]
@oleg-nenashev Please review the latest state of this PR. Would you also be so kind to answer this?

> Any reason to prefer @Nullable over @CheckForNull ?

**Comment by Smasherr** [comment]
Bump: any feedback?

**Comment by timja** [comment]
> @oleg-nenashev Please review the latest state of this PR. Would you also be so kind to answer this?
> 
> > Any reason to prefer `@Nullable` over `@CheckForNull`?

CheckForNull is normally preferred as it has better integration into static code analysis frameworks like spotbugs AFAIK

File: pom.xml [hunk 4]
```diff
@@ -237,6 +237,27 @@
       <version>1.2.7</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>1.31</version>
+      <scope>test</scope>
+      <!-- RequireUpperBoundDeps of maven-enforcer-plugin requires commons-beanutils:1.8.3
```
timja:
```suggestion
      <!-- TODO: remove after configuration-as-code 1.32 is released RequireUpperBoundDeps of maven-enforcer-plugin requires commons-beanutils:1.8.3
```

src/test/java/hudson/plugins/sonar/CasCTest.java
```diff
@@ -0,0 +1,47 @@
...
+import jenkins.model.GlobalConfiguration;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CasCTest extends RoundTripAbstractTest {
+
+    @Override
+    protected String configResource() {
+        return "CasCTest/config.yaml";
```
timja: Why not just put this test in the casc package? then you wouldn't need do this this?
the default file is called, configuration-as-code.yaml, you shouldn't really need to override it unless you're testing multiple combinations

Smasherr: Thanks for the hint

timja:
The build is failing because it's untrusted I assume:
`/home/travis/.local/bin/install: line 19: ARTIFACTORY_URL: unbound variable`

henryju:
@dbmeneses can you please take over this PR?

Smasherr:
Bump: merge it please

timja:
@Smasherr can you please resolve the conflicts?

Smasherr:
Would be cool to have this merged first: https://github.com/SonarSource/sonar-scanner-jenkins/pull/130

timja:
> Would be cool to have this merged first: #130

probably no harm in cherry-picking that into this PR as well

timja:
@henryju  / @dbmeneses could we get this reviewed / merged please? We have users asking about this often

src/test/resources/hudson/plugins/sonar/casc/configuration-as-code.yaml
```diff
@@ -0,0 +1,10 @@
+unclassified:
+  sonarglobalconfiguration:
+    buildWrapperEnabled: true
+    installations:
+      - name: "TEST"
+        credentialsId: "test-id"
+        serverUrl: "http://url:9000"
+        triggers:
+          skipScmCause: true
+          skipUpstreamCause: true
```
wouter-admiraal-sonarsource: Please add an empty new line here.

src/test/java/hudson/plugins/sonar/casc/CasCTest.java
```diff
@@ -0,0 +1,44 @@
...
+                .get(SonarGlobalConfiguration.class)
+                .getInstallations()[0];
+        assertThat(installation).isNotNull();
+    }
+
+    @Override
+    protected String stringInLogExpected() {
+        return "credentialsId = test-id";
+    }
+}
```
wouter-admiraal-sonarsource: I see we forgot to link to our contribution guidelines for this project. Sorry about that.

Could you please reformat this code so it follows our code style (config for Eclipse and IntelliJ available [here](https://github.com/SonarSource/sonar-developer-toolset#eclipse-configuration))? From the looks of it, I think only the indentation should be corrected.

timja:
> Nullable

See the discussion here, about `CheckForNull` and `Nullable`.
https://github.com/jenkinsci/configuration-as-code-plugin/pull/1053#issuecomment-529873362

The jenkins project tends to prefer `CheckForNull` as it has better integration with `spotbugs`. i.e. you must check a value for null if it could be null

src/test/resources/hudson/plugins/sonar/casc/configuration-as-code.yaml
timja:
```suggestion
          skipUpstreamCause: true

```

src/test/java/hudson/plugins/sonar/casc/CasCTest.java
timja: > I see we forgot to link to our contribution guidelines for this project. Sorry about that.
> 
> Could you please reformat this code so it follows our code style (config for Eclipse and IntelliJ available [here](https://github.com/SonarSource/sonar-developer-toolset#eclipse-configuration))? From the looks of it, I think only the indentation should be corrected.

Checkstyle? 🙂

henryju:
We are not using Spotbugs but SonarQube ;) From CasC point of view, does it make a difference if we switch to `@Nullable`?

timja:
> We are not using Spotbugs but SonarQube ;) From CasC point of view, does it make a difference if we switch to `@Nullable`?

Needs a code change in JCasC and a release, Oleg didn't seem terribly keen on Nullable but probably no big deal..

Smasherr:
Hey guys, nice to see finally some actions here. I've been on vacation and can do the requested changes to this PR tomorrow.

Smasherr:
@wouter-admiraal-sonarsource Hi, does it look better to you now?

wouter-admiraal-sonarsource:
> @wouter-admiraal-sonarsource Hi, does it look better to you now?

Perfect, thanks. I'll run the QA one more time from #132, and we'll finally merge. Thank you for your patience and contribution 👍 !

pom.xml
```diff
@@ -247,7 +247,21 @@
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>okhttp</artifactId>
       <version>3.7.0</version>
-</dependency>
+      <scope>test</scope>
```
henryju: Why changing this scope?

Smasherr: This dependency appeared in https://github.com/SonarSource/sonar-scanner-jenkins/commit/be10664e2519100efe6b356a62e6074b5cb5178a and as you see it is used there only `WaitForQualityGateStepTest`

Smasherr: Initially I wanted to improve the scope in a separate PR: https://github.com/SonarSource/sonar-scanner-jenkins/pull/130 but after a discussion https://github.com/SonarSource/sonar-scanner-jenkins/pull/124#issuecomment-545331715 we decided to include it in here.

Smasherr: I believe it was actually MEANT to be in test scope, especially if you see that it was added after the comment `<!-- Test deps -->`  https://github.com/SonarSource/sonar-scanner-jenkins/blob/master/pom.xml#L186

Smasherr:
@wouter-admiraal-sonarsource You're welcome! Would you be so kind and to do a release with this change anytime soon?

pom.xml
henryju: LGTM @tom-vanbraband-sonarsource do you confirm?

tom-vanbraband-sonarsource: Yes, that's my mistake. Thanks for fixing that @Smasherr

wouter-admiraal-sonarsource: So, this turns out to be incorrect, as it's used by some other dependency. I'll update this in #132 directly, and merge once the QA passes.

Smasherr:
Status?

nettworks-tooling:
Any news on this topic? Would be very cool if this would be merged soon as we spot issues at this point on our automated Jenkins master setup.

wouter-admiraal-sonarsource:
This was merged in #132. Thank you all for your help 👍

Smasherr:
@wouter-admiraal-sonarsource good news, thanks! Now it would be great to have this in a release any time soon :-)

Wurden:
I would like to add the it not only fails when serverAuthenticationToken is missing, it fails if any possible config parameter is missing, so for a working JasC you need to include everything even though values are empty. Most other plugins have default and does not fail when config parameters are empty.

`          sonarGlobalConfiguration:
            installations:
            - name: "url-to-sonarqube.com"
              serverUrl: "https://url-to-sonarqube.com/"
              serverAuthenticationToken: "" # Although we don't need any of below parameters, the Sonar plugin requires them to be present
              credentialsId: ""
              webhookSecretId: ""
              mojoVersion: ""
              additionalProperties: ""
              additionalAnalysisProperties: ""
              triggers:
                envVar: ""
                skipScmCause: false
                skipUpstreamCause: false`

Smasherr:
@Wurden The result of this change is that only 4 parameters remain required: name, serverUrl and triggers. The latter will become optional after #136 is merged

PR #129, opened 2019-10-16
refactor validate webhook to improve coverage

tom-vanbraband-sonarsource:


src/main/java/hudson/plugins/sonar/JenkinsSupplier.java
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package hudson.plugins.sonar;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+import jenkins.model.Jenkins;
+
+public class JenkinsSupplier implements Supplier<Jenkins> {
```
janos-ss: How about just using an inline supplier and dropping this class?

src/main/java/hudson/plugins/sonar/SonarGlobalConfiguration.java
```diff
@@ -145,36 +155,27 @@ public static SonarGlobalConfiguration get() {
...
-    return new StandardListBoxModel()
-      .includeEmptyValue()
-      .includeMatchingAs(
-        ACL.SYSTEM,
-        Jenkins.getInstance(),
-        StringCredentials.class,
-        Collections.emptyList(),
-        CredentialsMatchers.always()
-      );
+    return constructListBoxModel(credentialsId);
```
janos-ss: Nitpick: I think "create" is just a more common word for the pattern than "construct".

And definitely good to not duplicate so much code! (I should have caught earlier.)

PR #133, opened 2019-10-28
Update QA axis to latest LTS

henryju:


henryju:
Is not enough to fix the QA. Closing.

PR #144, opened 2020-07-01
Rename Quality Gate status in Jenkins job page

kumbasar:
This merge simply renames the Quality Gate status as follows:

- `OK` to `Passed`
- `ERROR` to `Failed`
- `WARN` to `Warning`
-  All others or empty to `N/A`

Which will be the same as the SonarQube ones. Also, they are more readable for the end-users. 

**Example screenshot**:
image


mark-rekveld-sonarsource:
This will be included by #149

PR #152, opened 2020-07-29
Fix QA

mark-rekveld-sonarsource:


its/pom.xml
```diff
@@ -17,6 +17,10 @@
 
   <inceptionYear>2013</inceptionYear>
 
+  <properties>
+    <sonar-ws.version>7.8</sonar-ws.version>
```
dbmeneses, line 21: maybe take the opportunity to update it to the LTS 7.9?

PR #153, opened 2020-08-10
SONARJNKNS-313 - Check for null on plain text secrets.

mark-rekveld-sonarsource:


src/main/java/hudson/plugins/sonar/GlobalCredentialMigrator.java
```diff
@@ -46,14 +47,19 @@ StandardCredentials migrate(@Nonnull String token) {
...
-        Jenkins.getInstance(),
+        Jenkins.getInstanceOrNull(),
         ACL.SYSTEM,
         (DomainRequirement) null),
       CredentialsMatchers.always());
 
     return allStringCredentials
       .stream()
-      .filter(cred -> cred.getSecret().getPlainText().equals(token))
+      // The plaintext secret should not be nullable but a NPE can still be thrown, SONARJNKNS-313, Optional wrapping
```
dbmeneses, line 57: Just my opinion, I think it would be easier to read something like 

`allStringCredentials.stream()`
`.filter(Objects::nonNull)`
`.filter(cred -> cred.getSecret().getPlainText().equals(token))`
`.findAny()`.`

mark-rekveld-sonarsource: Its may not be the cred that is null but the secret or the plaintext within the secret.

The funny thing is that all the APIs are annotated with `@NonNull` which by the exception listed in the issue is not true.

That is why I opted to treat the credentials as nullable as well as the secret it provides and the plaintext field the secret provides.

dbmeneses: ok, got it

PR #155, opened 2020-08-18
SONARJNKNS-320 - waitForQualityGate could get stuck

mark-rekveld-sonarsource:


src/main/java/org/sonarsource/scanner/jenkins/pipeline/SonarQubeWebHook.java
```diff
@@ -41,6 +45,7 @@
 @Extension
 public class SonarQubeWebHook implements UnprotectedRootAction {
   private static final Logger LOGGER = Logger.getLogger(SonarQubeWebHook.class.getName());
+  private final Cache<String, WebhookEvent> eventCache = CacheBuilder.newBuilder().expireAfterWrite(2, TimeUnit.HOURS).build();
```
mark-rekveld-sonarsource, line 48: The 2 hour window is a "magic" window. Seems like a large enough window that is not too large to negatively impact memory usage. I'm open to improvements here if you know of any.

PR #157, opened 2020-08-24
SONARJNKNS-309 - Don't warn on missing Analysis

mark-rekveld-sonarsource:


src/main/java/hudson/plugins/sonar/client/SQProjectResolver.java
```diff
@@ -79,6 +80,13 @@ public ProjectInformation resolve(@Nullable String serverUrl, @Nullable String p
 
       return projectInfo;
 
+    } catch (HttpException e) {
+      if (e.code() == 404) {
+        Logger.LOG.log(Level.FINE, "Error fetching project information: {0}", e.getMessage());
```
dbmeneses, line 85: Fine? is this really going to get used?

src/main/java/hudson/plugins/sonar/action/SonarProjectIconAction.java
```diff
@@ -44,7 +44,7 @@ public SonarProjectIconAction(SonarAnalysisAction buildInfo) {
 
   @Override
   public String getIconFileName() {
-    PluginWrapper wrapper = Jenkins.getInstance().getPluginManager()
+    PluginWrapper wrapper = Jenkins.getInstanceOrNull().getPluginManager()
```
dbmeneses, line 47: sounds like a NPE could happen here?

mark-rekveld-sonarsource: > sounds like a NPE could happen here?

In theory yes, but just like the `getInstance()` but this hides the fact the a NPE could get thrown.
In any case, the NPE is only thrown if Jenkins in not running, like in test cases, so I think we can safely ignore this warning in QG. One of the code bugs in Jenkins if you ask me with using static wiring.

src/main/java/hudson/plugins/sonar/client/SQProjectResolver.java
mark-rekveld-sonarsource: > Fine? is this really going to get used?

The main issue was that logs got polluted by these error messages if the analysis was not found.
This way we can hide the log messages from the logs by default, but allow us to instruct the user to increase the log level in case of issues that would benefit from this 'debug' entry.

Having typed al this, I'm open to just keeping it on 'info' or just dropping it completely.

dbmeneses: I honestly don't know how easy it is to enable fine, and if that's something often done. I'd assume fine is only used during for debugging while developing. Maybe debug is actually used in production? Really don't know but I agree that not showing in info is probably a good idea.

mark-rekveld-sonarsource: Jenkins has a UI for configuring a logger so this would be 'easy' enough to do, but I'm in the dark if this is something used often or not, it's a guestimate on my end.

PR #158, opened 2020-08-25
its: update Jenkins acceptance-test-harness (ATH)

thomasgl-orange:
Update ATH from 1.65 (+ a few patched classes) to ~1.74~ 1.75 (latest release). It enables running integration tests with more recent Jenkins versions (I've been trying 2.235.5, latest LTS). Without this upgrade, all the integration tests were failing, because the old ATH is not able to control the refreshed plugins manager UI.


mark-rekveld-sonarsource:
I have copied this PR to #159 to also run the QA build, and they don't seem to like to run with Jenkins version 2.176.4

thomasgl-orange:
@mark-clements-sonarsource: You're right, it seems ATH 1.73+ is not backward-compatible with Jenkins prior to 2.189, because of this change:
https://github.com/jenkinsci/acceptance-test-harness/pull/566
It is now waiting for a log message after installing plugins ("*Completed installation of X plugins*"), but this message has only been added in 2.189, via jenkinsci/jenkins@5169f14019, there:
https://github.com/jenkinsci/jenkins/pull/4124

Because of that, the tests end up in timeout...

I will try to submit an ATH PR to fix that (I have no idea whether 2.176.x is still a supported target for ATH though; it should in my opinion because 2.176.x still has its update-center available, but who knows).

thomasgl-orange:
I've opened jenkinsci/acceptance-test-harness#602, so let's wait and see what happens there. It fixes the issue on a few selected IT tests I've tried with 2.176.4 (I did not run the complete tests suite, it takes too much time/resources on my old laptop).

Of course, an alternative would be to stop supporting 2.176.x in this plugin. There are not that many users of the 2.11 version on 2.176.x already, and probably not all of them would upgrade to 2.12:
https://stats.jenkins.io/pluginversions/sonar.html
Also, I assume the 2.176 update center would disappear anyway once the next LTS (2.149.x) gets released (I think the update center is published only for the few last LTS baselines).

thomasgl-orange:
ATH 1.75 has been released, it includes jenkinsci/acceptance-test-harness#602.
I have push-forced on this PR to use 1.75 rather than 1.74, hopefully it will fix the ITs regressions when using Jenkins 2.176.4.

@mark-clements-sonarsource, could you update #159, to run a new QA build?

mark-rekveld-sonarsource:
I updated the QA build PR to see what will happen.

mark-rekveld-sonarsource:
The fix works, the QA builds are green. So I'll close this PR and request an internal review of #159

PR #167, opened 2020-10-16
Add a timeout option to the HttpClient to fix a bug

sprutner:
Naive first pass at introducing a default timeout
to HttpClient.java to fix a bug where the request will hang
on a non-response from a call to /api/server/version from
the Jenkins Master. The default timeout from HttpConnector is
a very long 30s to connect and 60s to read.

I couldn't figure out how to test this locally, but if introducing a change like this seems viable and there is a local development type doc I can reference I can clean it up a bit to fit that. I've never written Jenkins plugins, but we use this one and ran into an issue today with it when the master couldn't talk to /api/server/version. Not sure why it even needed to reach that, and when we used /etc/hosts to just route that request to a failed connection that solved our issue. I figure I should try to solve this problem for other people by making an upstream change.

### DIFF
--- a/src/main/java/hudson/plugins/sonar/client/HttpClient.java
+++ b/src/main/java/hudson/plugins/sonar/client/HttpClient.java
@@ -26,10 +26,14 @@ import org.sonarqube.ws.client.HttpConnector;
 import org.sonarqube.ws.client.WsResponse;
 
 public class HttpClient {
-  public String getHttp(String url, @Nullable String token) {
+  public String getHttp(String url, @Nullable String token, @Nullable Float timeout_ms) {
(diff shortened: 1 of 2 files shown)


mark-rekveld-sonarsource:
Hi @sprutner 

Thanks for the contribution.
Having a timeout seems like a good thing to have  especially with the call to `/api/server/version` that is executed from multiple places. But hardcoding it may have a negative impact for users that have slower SonarQube instances.

I did create https://jira.sonarsource.com/browse/SONARJNKNS-327 to address the need for timeouts where it can be configured allowing the timeout to be customized per user needs.

Cheers,
Mark

PR #168, opened 2020-10-21
Remove use of 'resetData' in ITs

dbmeneses:


julienlancelot:
Do not forget to upgrade orchestrator to 3.31.0.2646

its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java
```diff
@@ -122,7 +124,11 @@ public void setUp() {
 
   @After
   public void cleanup() {
+    if (!wsClient.projects().search(new SearchRequest().setProjects(Collections.singletonList(MVN_PROJECT_KEY))).getComponentsList().isEmpty()) {
```
julienlancelot, line 127: There's another interesting way to replace the resetDat() => it's to do a bulk delete of all projects.
See how the languages team has do it for instance here : https://github.com/SonarSource/sonar-dotnet/blob/98c0671a0b71ec36f1fc7a237aa51a7b6e5f57d7/its/src/test/java/com/sonar/it/shared/TestUtils.java#L187

julienlancelot (approved):
Please upgrade orchestrator, and feel free to apply my comment on the way to purge all projects at once.

its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java
dbmeneses: nice, I'll use that

PR #169, opened 2020-10-27
SC-2555 Mark pipeline as unstable when running Java < 11

tom-vanbraband-sonarsource:
The SonarCloud scanner logs a warning message when the scanner is executed with a Java version under 11. This change will mark the pipeline as unstable if it can find that log message. 

We saw from data that mainly Jenkins users are not making the switch to Java 11 so that's why we decided to focus on Jenkins specifically.

### DIFF
--- a/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
+++ b/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
@@ -24,6 +24,7 @@ import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.model.Action;
 import hudson.model.Actionable;
+import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.plugins.sonar.SonarInstallation;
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
```diff
@@ -116,8 +117,17 @@ public static SonarAnalysisAction addBuildInfoTo(Run<?, ?> build, TaskListener l
     boolean skippedIfNoBuild)
     throws IOException, InterruptedException {
     SonarAnalysisAction buildInfo = createSonarAnalysisAction(sonarInstallation, credentialId, build, listener);
-    Properties reportTask = extractReportTask(listener, workspace);
 
+    boolean java11Warning = build.getLog(Integer.MAX_VALUE).stream()
+      .anyMatch(line -> line.contains("Please update to at least Java 11."));
+
+    Result result = build.getResult();
+    if (java11Warning && (result == null || !result.equals(Result.FAILURE))) {
```
mark-rekveld-sonarsource, line 125: You could simply this to `java11Warning && Result.FAILURE.equals(result)`

mark-rekveld-sonarsource (changes requested):
The solution looks good. You do seem to have some build issues to fix.

tom-vanbraband-sonarsource:
@mark-rekveld-sonarsource the ITs are failing with the message:
```
java.lang.IllegalArgumentException: workflow-aggregator plugin not installed
```

I don't see how my change could have triggered that failure, do you have any idea what the cause might be?

mark-rekveld-sonarsource:
> I don't see how my change could have triggered that failure, do you have any idea what the cause might be?

I looked at the junit report for the test run and it seems that there are dependencies of the workflow-aggregator plugin that are not loading correctly. Looks like someone in Jenkins land published a new version of the plugin that is breaking our tests.

I have not yet found a solution to pinning plugin versions used in the tests

mark-rekveld-sonarsource:
@tom-vanbraband-sonarsource I committed a Jenkins version bump since I suspect this will fix the build. A more specific dependent plugin versions spec didn't get me a green build locally, so we are stuck with bumping Jenkins itself.

tom-vanbraband-sonarsource:
@mark-rekveld-sonarsource thanks a lot! This PR is ready to be merged and released from my side, let me know if I can help with anything!

mark-rekveld-sonarsource (approved):

PR #175, opened 2020-12-29
Remove tabular markup from forms

timja:
Following https://www.jenkins.io/doc/developer/views/table-to-div-migration/

See https://www.jenkins.io/blog/2020/11/10/major-changes-in-weekly-releases/

I'm not sure if this is causing an issue but I'm working through all the plugins that appear with tabular markup in https://issues.jenkins.io/browse/JENKINS-64341

### DIFF
--- a/src/test/java/hudson/plugins/sonar/SonarTestCase.java
+++ b/src/test/java/hudson/plugins/sonar/SonarTestCase.java
@@ -76,8 +76,12 @@ public abstract class SonarTestCase {
   protected Maven.MavenInstallation configureDefaultMaven() throws Exception {
     File mvn = new File(getClass().getResource("SonarTestCase/maven/bin/mvn").toURI().getPath());
     if (!Functions.isWindows()) {
-      // noinspection OctalInteger
-      GNUCLibrary.LIBC.chmod(mvn.getPath(), 0755);
+      try {
+        // noinspection OctalInteger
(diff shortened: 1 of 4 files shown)


src/test/java/hudson/plugins/sonar/SonarTestCase.java
```diff
@@ -76,8 +76,12 @@
   protected Maven.MavenInstallation configureDefaultMaven() throws Exception {
     File mvn = new File(getClass().getResource("SonarTestCase/maven/bin/mvn").toURI().getPath());
     if (!Functions.isWindows()) {
-      // noinspection OctalInteger
-      GNUCLibrary.LIBC.chmod(mvn.getPath(), 0755);
+      try {
+        // noinspection OctalInteger
+        GNUCLibrary.LIBC.chmod(mvn.getPath(), 0755);
```
timja, line 81: see javadoc from GNUCLibrary:
> Not available on all platforms so the caller should recover gracefully in case of {@link LinkageError}

this currently fails on Mac M1, adding a try catch appears to be enough for tests to pass here fine

mark-rekveld-sonarsource:
Thanks for the contribution. I have approved the CI to continue.

timja:
all green ^^

mark-rekveld-sonarsource (approved):

mark-rekveld-sonarsource:
and merged! thanks again for your contribution!!

PR #178, opened 2021-01-16
Use GitHub as the source of documentation

zbynek:
Replaces https://github.com/jenkinsci/sonarqube-plugin/pull/9

The reason to use GitHub URL is that  https://plugins.jenkins.io/sonar/ can only pull information from the Wiki (which is in read only mode and will be turned off eventually) or GitHub (the preferred way).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@ SonarQube Scanner for Jenkins
 
 [![Build Status](https://travis-ci.org/SonarSource/sonar-scanner-jenkins.svg?branch=master)](https://travis-ci.org/SonarSource/sonar-scanner-jenkins) [![Quality Gate](https://next.sonarqube.com/sonarqube/api/project_badges/measure?project=org.jenkins-ci.plugins%3Asonar&metric=alert_status)](https://next.sonarqube.com/sonarqube/dashboard?id=org.jenkins-ci.plugins%3Asonar)
 
+This plugin allow easy integration of [SonarQube™](https://www.sonarqube.org/), the open source platform for Continuous Inspection of code quality.
+
 Documentation: http://redirect.sonarsource.com/plugins/jenkins.html
 
 Issue tracking: https://jira.sonarsource.com/browse/SONARJNKNS
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
   <packaging>hpi</packaging>
 
   <name>SonarQube Scanner for Jenkins</name>
-  <url>http://redirect.sonarsource.com/plugins/jenkins.html</url>
+  <url>https://github.com/jenkinsci/sonarqube-plugin</url>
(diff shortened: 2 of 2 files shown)


mark-rekveld-sonarsource:
Hi @zbynek,

We prefer the use of the redirect service (redirect.sonarsource.com) so we can change documentation (including links to them) without having to release a new version. The use of this redirect service also allows older versions of the plugin to directed the correct documentation page.

Can you clarify why you would like to see the documentation url changed to point to Github?

Cheers,
Mark

zbynek:
Hi,

so the URL in the pom file is not shown to users directly (assuming they installed it using the plugin manager and didn't compile & upload it  themselves). Instead Jenkins communicates https://plugins.jenkins.io/sonar as the public plugin URL and the plugin site populates that page using the pom url.
* if the pom URL points to GitHub repo within Jenkins org, it will use GitHub API to render the readme file
* if it points to Jenkins wiki, it will use the Wiki HTML output
* if it points anywhere else, it will just show a link

Also note that the update center is currently overriding the URL from pom: https://github.com/jenkins-infra/update-center2/blob/update-center2-3.4.5/resources/wiki-overrides.properties#L54

So effectively Jenkins Wiki is used for Sonar documentation (it's just a short page with a link to the actual docs). Since Jenkins WIki is obsolete, there are two options for SonarQube docs:

1) use GitHub instead of the Wiki; this would mean just a brief  readme file that explains the purpose of the plugin and links to the actual docs (what I tried to achieve in this PR). Changes in the readme would be visible on plugins.jenkins.io without a plugin release, but for the initial switch of the URLs a release would be needed. The readme can be any .md or .adoc file in the rpository, not necessarily the same as the default GH readme.

2) keep just the redirect link in pom -- that way the plugin page https://plugins.jenkins.io/sonar  will not show any summary, just the link to the external documentation.

Please let me know which of these options sounds better to you.

Cheers,
Zbynek

mark-rekveld-sonarsource:
Hi @zbynek 

Thanks for the clarification. I approved the CI build to continue as I think your change is more valuable for users to get information and still be able to get directed to our documentation if needed.

Just one thought; Wouldn't it be easier to use the URL element within SCM as this already points to the Github repository?

Cheers,
Mark

zbynek:
Hi Mark, actually the update center uses the SCM URL as fallback, but that was only implemented recently and most plugins define the documentation URL explicitly, even if it's the same (see other PRs in the [docs migration project](https://github.com/orgs/jenkinsci/projects/3)). Also I'm not sure if the plugin center is the only part of Jenkins ecosystem that extracts the URL from POM.

mark-rekveld-sonarsource (approved):

mark-rekveld-sonarsource:
Thanks @zbynek for the added details, the PR has been merged not sure when the next release is planned.

PR #181, opened 2021-04-07
Stabilize QA

michal-duda-sonarsource:


wouter-admiraal-sonarsource (commented):
I see this comes from https://github.com/takari/maven-wrapper. I guess it's all OK, but I'm surprised that we commit the `.mvn/wrapper/maven-wrapper.jar` and `.mvn/wrapper/maven-wrapper.properties` files. Is this expected?

michal-duda-sonarsource:
I added the `.jar` file to `.gitignore` but the `.properties` file is needed if we want to fix on a specific maven version, which to me makes sense to keep the builds reproducible.

wouter-admiraal-sonarsource (approved):
To be clear, my question was if committing the wrapper was expected. I wasn't necessarily challenging the fact that it was committed or not. I'm just not familiar enough with the topic (Googl'ing showed people aren't aligned on the topic either, so I just wanted to double check it wasn't a `git add .` mistake).

Anyway, LGTM 👍 . Thanks for taking care of this.

PR #185, opened 2021-04-27
BUILD Use latest build docker image

malena-ebert-sonarsource:


tobias-trabelsi-sonarsource (changes requested):
Missing wget dependency in Dockerfile

PR #190, opened 2021-04-30
Fix MsBuild version in ITs

jacek-poreda-sonarsource:


lukasz-jarocki-sonarsource:
Code looks good. Do you think it is worth running it on Windows?

lukasz-jarocki-sonarsource (approved):

jacek-poreda-sonarsource:
I'm pretty sure it will fail on Windows, also it is not trivial to set it up. I would handle those tests when we add windows VM as part of the CI.

lukasz-jarocki-sonarsource:
Not sure what happened to my comment. Anyway. I did run at lest `mvn test` and it is green.

PR #196, opened 2021-06-18
Update jenkins tests to use latest LTS and compatible plugins

jacek-poreda-sonarsource:


dbmeneses:
Looks good, thanks for handling this.
You didn't add the unit test for the case when webhook is invalid, right?

dbmeneses (approved):

jacek-poreda-sonarsource:
There were already test for that, I've just added assertions for messages.

PR #198, opened 2021-07-02
NO-JIRA bumping jenkins version that we use for ITs

lukasz-jarocki-sonarsource:
This is the version that Jenkins recommends 

> Do not use versions no longer supported by the update center, which is currently anything older than 2.238 for weekly releases, and 2.222.4 for LTS releases.


jacek-poreda-sonarsource (approved):
LGTM 👍

PR #199, opened 2021-08-02
JENKINS-65091 Improve code coverage

wouter-admiraal-sonarsource:


src/main/java/hudson/plugins/sonar/action/SonarCacheAction.java
```diff
@@ -105,6 +104,12 @@ static boolean isEntryValid(@Nullable ProjectInformation cached, long lastBuild)
     return false;
   }
 
+  @VisibleForTesting
+  protected void cacheProjectInfo(List<ProjectInformation> list, Long time) {
+    lastProjInfo = list;
+    lastRequest = time;
+  }
+
```
wouter-admiraal-sonarsource, line 112: I'm using this workaround so I don't have to mock `System.currentTimeMillis()` in the unit tests. Not sure if there's a better way, though...

src/test/java/hudson/plugins/sonar/action/SonarCacheActionTest.java
```diff
@@ -90,18 +93,34 @@ public void testResolveUsingInstallationUrl() {
...
-    List<ProjectInformation> projs = cache.get(resolver, 0, Collections.singletonList(analysis), run);
-    List<ProjectInformation> projs2 = cache.get(resolver, 0, Collections.singletonList(analysis), run);
+    when(resolver.resolve(analysis1.getServerUrl(), analysis1.getUrl(), Objects.requireNonNull(analysis1.getCeTaskId()), analysis1.getInstallationName(), run)).thenReturn(mocked1);
+    when(resolver.resolve(analysis2.getServerUrl(), analysis2.getUrl(), Objects.requireNonNull(analysis2.getCeTaskId()), analysis2.getInstallationName(), run)).thenReturn(mocked2);
+
+    List<ProjectInformation> projs = cache.get(resolver, 0, Collections.singletonList(analysis1), run);
+    // Calling it again in quick succession, even with more analyses, should still return the cached value.
+    List<ProjectInformation> projs2 = cache.get(resolver, 0, Arrays.asList(analysis1, analysis2), run);
 
-    assertThat(projs).isEqualTo(projs2);
```
wouter-admiraal-sonarsource, line 101: This assertion was misleading. It's always the same object, regardless of the fact if you hit the cache or not. It's the size that can tell whether the cache was hit or not.

jacek-poreda-sonarsource (approved):
LGTM

PR #201, opened 2021-08-08
[JENKINS-66308] Prepare SonarQube Scanner for core Guava upgrade

basil:
See [JENKINS-66308](https://issues.jenkins.io/browse/JENKINS-66308) and [JENKINS-65988](https://issues.jenkins.io/browse/JENKINS-65988). Jenkins core is using [Guava 11.0.1](https://github.com/google/guava/releases/tag/v11.0.1), which was released on January 9, 2012. Jenkins core would like to upgrade to [Guava 30.1.1](https://github.com/google/guava/releases/tag/v30.1.1), which was released on March 19, 2021. Plugins must be prepared to be compatible with both Guava 11.0.1 and Guava 30.1.1 in advance of this core transition.

In particular, this plugin has been identified as using the `com.google.common.cache.Cache#get(K key)` API, which has been removed between Guava [11.0.1](https://guava.dev/releases/11.0.1/api/docs/com/google/common/cache/Cache.html) and [latest](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/cache/Cache.html). The removal took place in Guava 12.0.

To facilitate the Jenkins core transition, this plugin must be prepared and released such that it works with both Guava 11.0.1 and latest. The recommendation is for plugins to migrate from Guava's cache to [Caffeine](https://github.com/ben-manes/caffeine) via the Jenkins [Caffeine API](https://plugins.jenkins.io/caffeine-api/) plugin.

For the most part, the migration is a trivial matter of replacing imports of `com.google.common.cache.Cache` with imports of `com.github.benmanes.caffeine.cache.Cache` and imports of `com.google.common.cache.CacheBuilder` with imports of `com.github.benmanes.caffeine.cache.Caffeine`. This has already been done in other plugins in the following PRs:

- jenkinsci/active-directory-plugin#109
- jenkinsci/azure-ad-plugin#125
- jenkinsci/configuration-as-code-plugin#1596
- jenkinsci/kubernetes-plugin#991
- jenkinsci/lockable-resources-plugin#242
- jenkinsci/role-strategy-plugin#150

PR #204, opened 2021-08-26
Build metadata updates

basil:
The remaining build metadata updates from #201 for this plugin to remain in the status quo for Jenkins plugins.


belen-pruvost-sonarsource:
@dbmeneses is this a change you could assess? 🙏

basil:
The incrementals subsystem, which is widely used throughout the Jenkins plugin ecosystem, is documented in [JEP-305](https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc) and [incrementals-tools](https://github.com/jenkinsci/incrementals-tools).

wouter-admiraal-sonarsource:
OK, I've read the [announcement on the Jenkins blog](https://www.jenkins.io/blog/2018/05/15/incremental-deployment/) regarding incremental builds, as well as the [JEP-305 spec](https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc). It all seems good to me. Furthermore, build passes, QA passes... And this was introduced to the Jenkins ecosystem _in 2018_. So we could say we're a bit late to this party.

I'm going to approve this. Thanks @basil for your help and patience 👍 .

wouter-admiraal-sonarsource (approved):

PR #206, opened 2021-08-30
SONARJNKNS-331 IT failing on setting up Scanner for MSBuild

wouter-admiraal-sonarsource:
The IT tries to setup the oldest released version of the Scanner for MSBuild. It does this by using a select list in the administration section. However, this list is only showing 2 versions, and the oldest released version is not part of it:

Screenshot 2021-08-30 at 15 49 08

Hence, the IT fails, not finding the option in the list.

I don't understand where this comes from, but in order to stabilize the QA, I've made the ITs select only the 1st two options, which corresponds to the current behavior of our plugin (which might be a regression, BTW, but I couldn't pin-point the commit that introduced this change).

### DIFF
--- a/its/src/test/java/com/sonar/it/jenkins/ScannerSupportedVersionProvider.java
+++ b/its/src/test/java/com/sonar/it/jenkins/ScannerSupportedVersionProvider.java
@@ -40,14 +40,27 @@ public class ScannerSupportedVersionProvider {
    */
   public String getEarliestSupportedVersion(String scannerNameRepo) {
     try {
-      HttpUrl.Builder httpBuilder = HttpUrl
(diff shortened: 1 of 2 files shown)


jacek-poreda-sonarsource (approved):

timja:
@wouter-admiraal-sonarsource it comes from https://github.com/jenkins-infra/crawler/pull/108

by @mickael-caro-sonarsource

wouter-admiraal-sonarsource:
Thanks @timja, 👍 , I added the info to the [ticket](https://jira.sonarsource.com/browse/SONARJNKNS-331).

timja:
There's questions on that PR on whether it should be reverted / adjusted, it would be good if you could take a look

wouter-admiraal-sonarsource:
Yes, we're discussing it internally. If we revert the other change, we'll revert this IT update as well.

PR #218, opened 2022-04-13
BUILD-1446 Add sbom to the build

tomverin:
Update parent pom to enable cycloneDX plugin
Add sbom to the build

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,7 @@
     <gitRepositoryName>sonar-scanner-jenkins</gitRepositoryName>
     <version.artifactory.plugin>2.6.1</version.artifactory.plugin>
     <jenkins.version>2.222.4</jenkins.version>
+    <version.cyclonedx.plugin>2.5.3</version.cyclonedx.plugin>
     <java.level>8</java.level>
     <findbugs.failOnError>false</findbugs.failOnError>
     <!-- Disable parallel tests on Travis -->
@@ -70,7 +71,7 @@
     <license.years>${project.inceptionYear}-2022</license.years>
     <license.mailto>mailto:info AT sonarsource DOT com</license.mailto>
 
-    <artifactsToPublish>${project.groupId}:${project.artifactId}:hpi</artifactsToPublish>
+    <artifactsToPublish>${project.groupId}:${project.artifactId}:hpi${project.groupId}:${project.artifactId}:json:cyclonedx</artifactsToPublish>
 
     <maven.compiler.release>8</maven.compiler.release>
   </properties>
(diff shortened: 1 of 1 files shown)


tomverin:
sbom is available
https://repox.jfrog.io/artifactory/sonarsource-public-qa/org/jenkins-ci/plugins/sonar/2.15.0.3695/sonar-2.15.0.3695-cyclonedx.json

tomverin:
sbom file available here https://repox.jfrog.io/artifactory/sonarsource-public-dev/org/jenkins-ci/plugins/sonar/2.15.0.3695/sonar-2.15.0.3695-cyclonedx.json

julien-carsique-sonarsource (approved):

pom.xml
```diff
@@ -70,7 +71,7 @@
     <license.years>${project.inceptionYear}-2022</license.years>
     <license.mailto>mailto:info AT sonarsource DOT com</license.mailto>
 
-    <artifactsToPublish>${project.groupId}:${project.artifactId}:hpi</artifactsToPublish>
+    <artifactsToPublish>${project.groupId}:${project.artifactId}:hpi${project.groupId}:${project.artifactId}:json:cyclonedx</artifactsToPublish>
```
wouter-admiraal-sonarsource, line 74: Sorry, I have no idea, but is this correct? Shouldn't there be a `,` or a `:` after `:hpi`?

pom.xml
```diff
@@ -54,6 +54,7 @@
     <gitRepositoryName>sonar-scanner-jenkins</gitRepositoryName>
     <version.artifactory.plugin>2.6.1</version.artifactory.plugin>
     <jenkins.version>2.222.4</jenkins.version>
+    <version.cyclonedx.plugin>2.5.3</version.cyclonedx.plugin>
```
wouter-admiraal-sonarsource, line 57: I'm guessing we're using this same property everywhere? If not, can we align it with the previous one and call it `cyclonedx.plugin.version` for consistency? Otherwise, leave it like this 👍 .

pom.xml
tomverin: good point, thanks

pom.xml
tomverin: Indeed this was factored in the parent pom

pom.xml
wouter-admiraal-sonarsource: Looking at our Ant Scanner, I think so. Missing an `,`.

pom.xml
tomverin: scratch that, this scanner does not use the parent pom so we have to re-declare everything for this one

PR #219, opened 2022-04-19
NO-JIRA fix build

aurelien-poscia-sonarsource:


.cirrus.yml
```diff
@@ -75,9 +75,9 @@ linux_qa_task:
     BROWSER: chrome
     matrix:
       - SQ_VERSION: LATEST_RELEASE[8.9]
-        JENKINS_VERSION: 2.263.1
+        JENKINS_VERSION: 2.277.1
       - SQ_VERSION: DEV
-        JENKINS_VERSION: 2.263.1 ### https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
+        JENKINS_VERSION: 2.277.1 ### https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
```
aurelien-poscia-sonarsource, line 80: I wonder what is the strategy here? Do we try to be compatible with oldest possible jenkins version? what is the purpose?

Shouldn't we bump to a more recent version to:
1. make sure we are actually compatible with recent versions of jenkins
2. avoid breaking again the build in few months?

Ideally I guess we should test against a range of versions containing the most recent jenkins' version

jacek-poreda-sonarsource: As you said, the strategy is to support a minimal stable version of Jenkins

jacek-poreda-sonarsource (approved):

PR #221, opened 2022-05-27
Fix nightly QA

wouter-admiraal-sonarsource:
The problem comes from incompatible plugin dependencies. This is likely one of the major reasons for all the instabilities we've had with our QA over the past months.

**Problem**
We install a fixed version of Jenkins (usually the LTS), but we always fetch the latest version of any plugin. This can cause issues, because:

- Some latest versions of plugins are not compatible with the running version of Jenkins, especially as a new LTS is published every 3 months.
- Some plugins have a shared dependency, but they might not require the same minimum version of that shared dependency. Jenkins will simply install the minimum version required by the first plugin that defines this dependency, which can fail the installation of another plugin that may require a higher version.

The problem is compounded by the fact this will fail "silently" (info only in the Jenkins logs), and the IT will continue its work, assuming the plugins were successfully installed. This is why they fail at stages like the plugin configuration screen, where expected options don't show up.

**Solution**
Pin the versions of all required plugins to something specific. If a dependency can cause a conflict due to different plugins requiring different versions, pin the version of this dependency to the most commonly compatible version and install it directly, don't treat it as a dependency. Pin everything, including Jenkins, to something that was working a few weeks ago, and keep it there for now.

This will hopefully stabilize our QA.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -74,10 +74,12 @@ linux_qa_task:
   env:
     BROWSER: chrome
     matrix:
+      # If updating the JENKINS_VERSION, make sure that all installed plugins are compatible. See com.sonar.it.jenkins.SonarPluginTest
+      # https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
       - SQ_VERSION: LATEST_RELEASE[8.9]
         JENKINS_VERSION: 2.277.1
       - SQ_VERSION: DEV
-        JENKINS_VERSION: 2.277.1 ### https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
+        JENKINS_VERSION: 2.277.1
   maven_cache:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
   qa_script:
--- a/its/src/test/java/com/sonar/it/jenkins/JenkinsUtils.java
+++ b/its/src/test/java/com/sonar/it/jenkins/JenkinsUtils.java
@@ -441,17 +441,6 @@ public class JenkinsUtils {
     });
   }
 
-  public void assertNoSonarPublisher(String jobName, File projectPath) {
-    try {
-      newFreestyleJobConfig(jobName, projectPath);
-      findElement(buttonByText("Add post-build action")).click();
-      assertNoElement(MAVEN_POST_BUILD_LABEL);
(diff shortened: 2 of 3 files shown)


jacek-poreda-sonarsource (approved):
Nice one, just for future reference, where did you check what plugins version to use?

wouter-admiraal-sonarsource:
> Nice one, just for future reference, where did you check what plugins version to use?

A bit of trial and error really 😅 . I went to the "releases" and "dependencies" screens for the major plugins we require (like [workflow-aggregator](https://plugins.jenkins.io/workflow-aggregator/)), and tried to figure it out (we cannot use the latest releases, because of [this unmaintained plugin](https://plugins.jenkins.io/filesystem_scm/) we need). The tricky part is transitive dependencies, like git-server.

Note that this is _not_ fault-proof. workflow-aggregator has a ton of dependencies, and Jenkins will happily try and install the latest version of some of them. At some point, one of these will probably break our QA again. In the long run, we need a better strategy for this (perhaps pinning _all_ dependencies, including transitive ones; but there are at least a dozen of these).

aurelien-poscia-sonarsource:
The review was merged will I have looking at it... but I have a few questions :)

Isn't the goal of this IT specifically to verify if a standard jenkins setup with latest dependencies would be compatible with the SQ scanners?

Fixing the version solves the instabilities of the build; at the same time, how do we assert that the scanner is still compatible with more recent versions of other plugins? also, what is the workflow to upgrade those dependencies? I'm afraid we would forever depend on a stable version of all those dependencies (no-one will think of updating them) and that this test will not verify that we remain compatible with more recent versions (but that might not be its goal?).

Also, I'm wondering whether some plugins are really needed. For instance _git-server_ or _filesystem_scm_. I would expect this test to "simply" install a jenkins, create a project and run our scanners. Why does it need to have a git server installed 🤔 ?

wouter-admiraal-sonarsource:
@aurelien-poscia-sonarsource You are absolutely right. I was thinking of pushing for a Team Topic to work on this, because this QA is -- IMHO -- completely broken. It does stuff we don't understand, it doesn't give us confidence in what we ship, some test cases are never run (they require Windows)... I think we need to rethink the QA entirely.

aurelien-poscia-sonarsource:
Thanks for your answer, it would indeed be great to have a deep look at this QA and make sure it's testing what we want and that it is maintainable :)

PR #224, opened 2022-07-08
Fixing QA

lukasz-jarocki-sonarsource:
Relatively big changes here:
1. Locking versions with [VERSION_OVERRIDES](https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/SUT-VERSIONS.md) env variable is more reliable than merely expressing preference of desired version inside @WithPlugin annotation. ✔️  
2. We had to migrate to 2.289.1 for a few reasons. One of the reason is expressed on the [Jenkins website ](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/) pretty clearly: 

> Do not use versions no longer supported by the update center, which is currently anything older than 2.291 for weekly releases, and 2.289.1 for LTS releases.

But our main reason to bump jenkins version was that transitive dependencies of Jenkins plugins didn't want to start on anything lower than 2.289.1 (there are ways around it, in example locking versions of transitive dependencies, but in our case they didn't help). 
And yet another reason to upgrade was that the Jenkins Update center for 2.771.1 started redirecting us to update center for 2.289.1 version (click [here](https://updates.jenkins-ci.org/update-center.json?version=2.277.1) to experience it). This could might have caused some issues when updating/installing plugins to their latest versions. ✔️ 

3. Some plugins in @WithPlugins annotation were removed - they were installed by Jenkins anyway when installing Sonar plugin (Jenkins is smart enough to install dependencies/plugins if they are needed to run "main" plugin). And less is more. ✔️ 

4. Locking version of git-server=1.9 is however the "meat" of this PR. If not locked, Jenkins installs git-server latest version [which requires](https://plugins.jenkins.io/git-server/#dependencies) at least 3.1.0 SSH Server plugin. However, when starting 2.289.1 instance (at least when using acceptance-test-harness ) the SSH Server 3.0.3 is installed before anything else runs. SSH Server is one of detached plugin and Jenkins [official LTS upgrade notes](https://www.jenkins.io/doc/upgrade-guide/2.289/#upgrading-to-jenkins-lts-2-289-1) tells us to watch for it. Now, the easy solution for this would be to upgrade SSH Server to 3.1.0 before installing git-server. But this can't be done, as upgrading SSH Server plugin requires restart of Jenkins, and restarting Jenkins right now from ITs is hard or next to impossible (I am happy to provide more details here if needed).

So in the end the solution here is to install git-server version compatible with default SSH Server 3.0.3 version and lock SSH Server 3.0.3 to make sure both of the plugins are compatible with eachother. ✔️ 

5. Last thing, completely unrelated to all of the points above was to update deleting of the projects, as the endpoint that we used for it has changed a little bit - providing future date to the endpoint is not possible anymore ✔️

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -77,9 +77,11 @@ linux_qa_task:
       # If updating the JENKINS_VERSION, make sure that all installed plugins are compatible. See com.sonar.it.jenkins.SonarPluginTest
       # https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/
       - SQ_VERSION: LATEST_RELEASE[8.9]
-        JENKINS_VERSION: 2.277.1
+        JENKINS_VERSION: 2.289.1
+        VERSION_OVERRIDES: git-server=1.9,sshd=3.0.3
       - SQ_VERSION: DEV
-        JENKINS_VERSION: 2.277.1
+        JENKINS_VERSION: 2.289.1
+        VERSION_OVERRIDES: git-server=1.9,sshd=3.0.3
   maven_cache:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
   qa_script:
--- a/its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java
+++ b/its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java
@@ -19,12 +19,6 @@
  */
 package com.sonar.it.jenkins;
 
-import static com.sonar.it.jenkins.JenkinsUtils.DEFAULT_SONARQUBE_INSTALLATION;
-import static java.util.Objects.requireNonNull;
-import static java.util.regex.Matcher.quoteReplacement;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.fail;
-
 import com.google.common.net.UrlEscapers;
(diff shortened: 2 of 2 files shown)


leo-geoffroy-sonarsource:
I had a look at the PR, and it looks good to me and run as expected in local
I wonder:
-  if a similar issue will happen again in the future. As I checked the tests, I felt that it was not straight forward to run, and I imagine the actual debugging of plugin versions may have been even more cumbersome. Would you be able to share the useful information that you used to achieve the result (matrix dependencies plugins or any meaningful resources) so it easy to come back to it if the problem reproduces?
I will validate the PR since everything else LGTM

lukasz-jarocki-sonarsource:
> I had a look at the PR, and it looks good to me and run as expected in local I wonder:
> 
>     * if a similar issue will happen again in the future. As I checked the tests, I felt that it was not straight forward to run, and I imagine the actual debugging of plugin versions may have been even more cumbersome. Would you be able to share the useful information that you used to achieve the result (matrix dependencies plugins or any meaningful resources) so it easy to come back to it if the problem reproduces?
>       I will validate the PR since everything else LGTM

Hey, yes Leo, this is what I have in mind also. I will share information on Confluence and leave a link on the PR to newly created article too. I will do it this week.

PR #226, opened 2022-09-02
Bump up version of jenkins and jenkins-ci test harness (selenium)

steve-marion-sonarsource:
The integration tests are failing on the CI because some plugins can't be installed anymore.


steve-marion-sonarsource:
The solution is not working on cirrus. Other works are being conducted that should fix the underlying issue.

PR #241, opened 2022-12-30
[NO-JIRA] Set a qa build environment supporting JAVA17

carminevassallo:
We have an issue when running our cron job related to a jre environment that cannot run our newest SQ. Therefore, we should upgrade to JAVA17 our QA environment


its/projects/abacus/pom.xml
```diff
@@ -54,6 +54,7 @@
     <sonar.pluginName>Abacus</sonar.pluginName>
     <sonar.pluginClass>org.sonar.plugins.abacus.AbacusPlugin</sonar.pluginClass>
     <sonar.scm.disabled>true</sonar.scm.disabled>
+    <argLine> --add-opens java.base/java.lang=ALL-UNNAMED</argLine>
```
carminevassallo, line 57: Shall we remove the extra space at the beginning?

matteo-mara-sonarsource: Yes totally!

matteo-mara-sonarsource (approved):
LGTM. The QA is fully green for both previous LTS and DEV versions.

PR #243, opened 2023-01-24
SONARJNKNS-357 Update license headers for 2023

wouter-admiraal-sonarsource:


wouter-admiraal-sonarsource:
Just realized I was missing the NOTICE.txt update here too. Added and re-pushed.

stanislavhh (approved):

PR #247, opened 2023-02-01
SONARJNKNS-358: Fix IT for Scanner MSBuild

antoine-vigneau-sonarsource:


antoine-vigneau-sonarsource:
@aurelien-poscia-sonarsource 

- I couldn't understand the plugin logic (not the test one) to only offer those 30 releases. It's weird, to the point that I feel it's implemented elsewhere, or provided by config. So I forced the test to pull 30 releases and take the latest and oldest from here (assuming this is the range the plugin offers too)
- `EARLIEST_JENKINS_INSTALLABLE_MSBUILD_VERSION`  is used elsewhere than the changed test, namely in `SonarPluginWindowsTest` but 1/ I think those tests also aim to get the oldest available version, otherwise it would have been a hardcoded one, and 2/ I don't think `SonarPluginWindowsTest` is executed at all in the CI, as they would fail (because forcing a way too old version `2.3.2.573` that the plugin won't offer). So this is why I safely set this value globally.

its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java
```diff
@@ -76,15 +76,18 @@ public void scan_project_sonarqube_scanner_v_3_3() {
 
   @Test
   public void scan_project_ms_build_net_core_scanner() {
-    MSBuildScannerInstallation.install(jenkins, EARLIEST_JENKINS_SUPPORTED_MS_BUILD_VERSION, false);
-    MSBuildScannerInstallation.install(jenkins, MS_BUILD_RECENT_VERSION, true);
+    if (!versionsAreDifferent(EARLIEST_JENKINS_INSTALLABLE_MSBUILD_VERSION, LATEST_JENKINS_MSBUILD_VERSION)) {
+      return;
```
aurelien-poscia-sonarsource, line 80: This is dangerous: in case the versions are the same, the test will succeed although it won't test anything.

Here you should throw an exception to make sure the test fails and an alarm is raised.

aurelien-poscia-sonarsource: You could solve the problem above + this one by replacing your code by:
`checkState(!EARLIEST.equals(LATEST), "Some description of the error");`

its/src/test/java/com/sonar/it/jenkins/utility/ScannerAvailableVersionsProvider.java
```diff
@@ -27,27 +27,49 @@
...
+    private String latest;
+    private String oldest;
+
+    public String getLatest() {
+      return latest;
+    }
+
+    public void setLatest(String latest) {
+      this.latest = latest;
+    }
```
aurelien-poscia-sonarsource, line 65: Instead of providing setters, this object could be immutable and have both version injected in the constructor. This will prevent any state mutation leading to unexpected situations.

aurelien-poscia-sonarsource:
> @aurelien-poscia-sonarsource
> 
> * I couldn't understand the plugin logic (not the test one) to only offer those 30 releases. It's weird, to the point that I feel it's implemented elsewhere, or provided by config. So I forced the test to pull 30 releases and take the latest and oldest from here (assuming this is the range the plugin offers too)
I had a look at this and here is how it works:
1. The [crawler](https://github.com/jenkins-infra/crawler/blob/master/sonarqubescannermsbuild.groovy) fetches the version from GitHub
2. It generates this [file](https://ftp.halifax.rwth-aachen.de/jenkins/updates/updates/hudson.plugins.sonar.MsBuildSonarQubeRunnerInstaller.json)
3. This file is in turn used to populate the dropdown

As the crawler does not set any preference for page size, it gets the default. So it's probably worth to do it the same thing here.


> * `EARLIEST_JENKINS_INSTALLABLE_MSBUILD_VERSION`  is used elsewhere than the changed test, namely in `SonarPluginWindowsTest` but 1/ I think those tests also aim to get the oldest available version, otherwise it would have been a hardcoded one, and 2/ I don't think `SonarPluginWindowsTest` is executed at all in the CI, as they would fail (because forcing a way too old version `2.3.2.573` that the plugin won't offer). So this is why I safely set this value globally.
You're right, `SonarPluginWindowsTest`, is only run in case the OS is windows. In the context of the CI, we start a linux container so it is never executed. P

its/src/test/java/com/sonar/it/jenkins/SonarPluginTestSuite.java
```diff
@@ -83,8 +81,11 @@ public static void setUpJenkins() {
       .build());
     DEFAULT_QUALITY_GATE_NAME = getDefaultQualityGateName();
 
-    EARLIEST_JENKINS_SUPPORTED_MS_BUILD_VERSION = SCANNER_VERSION_PROVIDER
-        .getEarliestSupportedVersion("sonar-scanner-msbuild");
+    ScannerAvailableVersionsProvider scannerProvider = new ScannerAvailableVersionsProvider();
+    ScannerAvailableVersionsProvider.ScannerAvailableVersions scannerVersions = scannerProvider.getScannerAvailableVersions("sonar-scanner-msbuild");
+
+    EARLIEST_JENKINS_INSTALLABLE_MSBUILD_VERSION = scannerVersions.getOldest();
+    LATEST_JENKINS_MSBUILD_VERSION = scannerVersions.getLatest();
```
aurelien-poscia-sonarsource, line 88: We should streamline variable names. What about:
* `OLDEST_INSTALLABLE_MSBUILD_VERSION`
* `LATEST_INSTALLABLE_MSBUILD_VERSION`

aurelien-poscia-sonarsource (commented):
Good improvement 👍 the test is now future proof 🎊 I raised a few comments, I'm happy to discuss them with you and/or pair to address them if you want so :)

its/src/test/java/com/sonar/it/jenkins/utility/ScannerAvailableVersionsProvider.java
```diff
@@ -27,27 +27,49 @@
...
     } catch (IOException | JSONException e) {
       throw new IllegalStateException(String.format("Could not fetch earliest supported version of %s", scannerNameRepo), e);
     }
+    ScannerAvailableVersions scannerAvailableVersions = new ScannerAvailableVersions();
+    scannerAvailableVersions.setLatest(versions.getJSONObject(0).getString("tag_name"));
+    scannerAvailableVersions.setOldest(versions.getJSONObject(versions.length() - 1).getString("tag_name"));
+    return scannerAvailableVersions;
+  }
+
+  public static class ScannerAvailableVersions {
```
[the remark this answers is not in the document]
antoine-vigneau-sonarsource: > Why not simply `MsBuildScannerVersions` (and `MsBuildScannerVersionsProvider`)
Because while we use this provider only for the MSBuild Scanner, it could be re-used for other scanners (parameter `scannerNameRepo`). And even any repository in fact (_that is not necessarily great though..._).
In fact, we might want to improve and also install the CLI Scanner with this mechanism in the future (not the case today).

its/src/test/java/com/sonar/it/jenkins/SonarPluginTest.java
```diff
@@ -76,15 +77,16 @@ public void scan_project_sonarqube_scanner_v_3_3() {
 
   @Test
   public void scan_project_ms_build_net_core_scanner() {
-    MSBuildScannerInstallation.install(jenkins, EARLIEST_JENKINS_SUPPORTED_MS_BUILD_VERSION, false);
-    MSBuildScannerInstallation.install(jenkins, MS_BUILD_RECENT_VERSION, true);
+    checkState(!OLDEST_INSTALLABLE_MSBUILD_VERSION.equals(LATEST_INSTALLABLE_MSBUILD_VERSION), "The test aims to run 2 different versions of the Scanner MSBuild, but they are not.");
```
aurelien-poscia-sonarsource, line 80: NP: is "run" accurate? should it be install?

its/src/test/java/com/sonar/it/jenkins/utility/ScannerAvailableVersionsProvider.java
```diff
@@ -27,27 +27,45 @@
...
+      versions.getJSONObject(versions.length() - 1).getString("tag_name")
+    );
+    return scannerAvailableVersions;
+  }
+
+  public static class ScannerAvailableVersions {
+    private String latestInstallableVersion;
+    private String oldestInstallableVersion;
+
+    public ScannerAvailableVersions(String latestInstallableVersion, String oldestInstallableVersion) {
```
aurelien-poscia-sonarsource, line 58: that's ultra NP, but I'd swap the argument as I feel more intuitive to have things in chronogical order

its/src/test/java/com/sonar/it/jenkins/utility/ScannerAvailableVersionsProvider.java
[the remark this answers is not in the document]
aurelien-poscia-sonarsource: good point indeed, the component could be reused for other scanners 👍

Actually, I edited my comment, firstly it was:
> Why not simply ScannerVersions (and ScannerVersionsProvider)

And I think this part still makes sense (or at least to replace `Available` with `Installable` for clarity/consistency).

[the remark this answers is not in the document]
antoine-vigneau-sonarsource: Ah ok got it, it makes sense to shorten those names indeed!

aurelien-poscia-sonarsource (approved):
LGTM 🚀 !!

PR #248, opened 2023-02-03
SONARJNKNS-360: Fix Quality Gate endpoint usage

antoine-vigneau-sonarsource:


its/pom.xml
```diff
@@ -18,7 +18,7 @@
   <inceptionYear>2013</inceptionYear>
 
   <properties>
-    <sonar-ws.version>8.9.0.43852</sonar-ws.version>
+    <sonar-ws.version>10.0.0.65447</sonar-ws.version>
```
aurelien-poscia-sonarsource, line 21: I'm afraid that as this artifact is not a release, it will be deleted in a few weeks. I guess the 8.9.0.43852 corresponds to the latest 8.9 LTS... You can probably use the build number of the 9.9 LTS here :)

antoine-vigneau-sonarsource: As discussed, using 9.9 won't solve the problem in itself. Though I'll do this, as it makes more sense, and fix the actual problem with another method (ie. doing the API call, not rely on sonar-ws)

aurelien-poscia-sonarsource (approved):
LGTM 🚀

aurelien-poscia-sonarsource:
Just a small comment for the future: the commit message could be more descriptive

antoine-vigneau-sonarsource:
> Just a small comment for the future: the commit message could be more descriptive

Thanks. That makes sense, I likely went too fast on this, my initial commit messages were better.

PR #253, opened 2023-05-01
[NO-JIRA] Update ITs openssl dependency

matteo-mara-sonarsource:
This openssl is breaking the sonar-scanner-jenkins build. Update with the new available version.


dimitris-kavvathas-sonarsource (approved):
LGTM

PR #258, opened 2023-08-04
SONARJNKNS-363 Remove version check

timja:
Please be aware that we are not actively looking for feature contributions. The truth is that it's extremely difficult for someone outside SonarSource to comply with our roadmap and expectations. Therefore, we typically only accept minor cosmetic changes and typo fixes. If you would like to see a new feature, please create a new thread in the forum ["Suggest new features"](https://community.sonarsource.com/c/suggestions/features).

With that in mind, if you would like to submit a code contribution, make sure that you adhere to the following guidelines and all tests are passing:

- [ ] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make

We are experiencing multiple hangs of our Jenkins instance a day recently, it happens when people visit a build page and the build refuses to load, eventually so many threads get blocked up that Jenkins restarts.

I took thread dumps which are attached below, uploading them to fastthread.io showed 90 threads blocked on:

[performanceData.7.output.tar.gz](https://github.com/SonarSource/sonar-scanner-jenkins/files/12261847/performanceData.7.output.tar.gz)

```
java.lang.Thread.State: BLOCKED (on object monitor)
at java.util.concurrent.ConcurrentHashMap.compute(java.base@11.0.20/ConcurrentHashMap.java:1923)
- waiting to lock <0x000000048cc315d0> (a java.util.concurrent.ConcurrentHashMap$Node)
at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2683)
at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2666)
at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:112)
at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:62)
at hudson.plugins.sonar.client.SQProjectResolver.resolve(SQProjectResolver.java:66)
at hudson.plugins.sonar.action.SonarCacheAction.get(SonarCacheAction.java:76)
at hudson.plugins.sonar.action.SonarCacheAction.get(SonarCacheAction.java:52)
at hudson.plugins.sonar.action.SonarProjectActionFactory.createProjectPage(SonarProjectActionFactory.java:118)
at hudson.plugins.sonar.action.SonarProjectActionFactory.createFor(SonarProjectActionFactory.java:84)
[...]

PR #265, opened 2023-09-01
SONARJNKNS-361 Configuration setting to use sonar.token

ericg138:
Adds new setting in the SonarQube server configuration

image

The setting is disabled by default to not break configurations with SQ < 10.0.

### DIFF
--- a/src/main/java/hudson/plugins/sonar/SonarInstallation.java
+++ b/src/main/java/hudson/plugins/sonar/SonarInstallation.java
@@ -35,6 +35,8 @@ import org.kohsuke.stapler.DataBoundConstructor;
 public class SonarInstallation implements Serializable {
 
   private static final long serialVersionUID = 1L;
+  static final String PROPERTY_SONAR_TOKEN = "sonar.token";
+  static final String PROPERTY_SONAR_LOGIN = "sonar.login";
 
   private final String name;
   private final String serverUrl;
@@ -69,6 +71,8 @@ public class SonarInstallation implements Serializable {
 
   private String[] split;
 
+  private final boolean useTokenProperty;
+
   /**
    * Maintained to retain compatibility
    * @deprecated since 2.9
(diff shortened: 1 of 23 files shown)


src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -256,6 +262,14 @@ public TriggersConfig getTriggers() {
     return triggers;
   }
 
+  public boolean useTokenProperty() {
```
timja, line 265: this should follow java bean conventions so it works fully with the https://plugins.jenkins.io/configuration-as-code plugin:

```suggestion
  public boolean isUseTokenProperty() {
```

src/main/java/hudson/plugins/sonar/SonarInstallation.java
```diff
@@ -92,7 +96,8 @@ public SonarInstallation(
     @CheckForNull String mojoVersion,
     @CheckForNull String additionalProperties,
     @CheckForNull String additionalAnalysisProperties,
-    @CheckForNull TriggersConfig triggers) {
+    @CheckForNull TriggersConfig triggers,
+    boolean useTokenProperty) {
```
timja, line 100: by adding a new non nullable parameter to the constructor this is a breaking change for configuration as code users and you've updated the config files to make it pass.

This may be okay and just calling it out here in case.

If you wanted to make it compatible you can set it via a setter annotated with `@DataBoundSetter`

timja (changes requested):

timja (commented):
another option would be to use the code that is being removed in https://github.com/SonarSource/sonar-scanner-jenkins/pull/258 to do a feature detection check based on the version and not make this configurable?

ericg138:
After further consideration, we will go with the solution to call SQ to get the version and set the parameter name accordingly.
Thank you @timja for the review.

PR #274, opened 2023-10-06
SONARJNKNS-366 Default token property when SQ version can't be fetched

ericg138:


jacek-poreda-sonarsource (approved):
Just small comment, otherwise LGTM 👍

PR #259, opened 2023-08-07
Save webhook results in action

timja:
Please be aware that we are not actively looking for feature contributions. The truth is that it's extremely difficult for someone outside SonarSource to comply with our roadmap and expectations. Therefore, we typically only accept minor cosmetic changes and typo fixes. If you would like to see a new feature, please create a new thread in the forum ["Suggest new features"](https://community.sonarsource.com/c/suggestions/features).

With that in mind, if you would like to submit a code contribution, make sure that you adhere to the following guidelines and all tests are passing:

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make

We are experiencing multiple hangs of our Jenkins instance a day recently, it happens when people visit a build page and the build refuses to load, eventually so many threads get blocked up that Jenkins restarts.

I took thread dumps which are attached below, uploading them to fastthread.io showed 90 threads blocked on:

[performanceData.7.output.tar.gz](https://github.com/SonarSource/sonar-scanner-jenkins/files/12261847/performanceData.7.output.tar.gz)

```
java.lang.Thread.State: BLOCKED (on object monitor)
at java.util.concurrent.ConcurrentHashMap.compute(java.base@11.0.20/ConcurrentHashMap.java:1923)
- waiting to lock <0x000000048cc315d0> (a java.util.concurrent.ConcurrentHashMap$Node)
at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2683)
at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2666)
at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:112)
at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:62)
at hudson.plugins.sonar.client.SQProjectResolver.resolve(SQProjectResolver.java:66)
at hudson.plugins.sonar.action.SonarCacheAction.get(SonarCacheAction.java:76)
at hudson.plugins.sonar.action.SonarCacheAction.get(SonarCacheAction.java:52)
at hudson.plugins.sonar.action.SonarProjectActionFactory.createProjectPage(SonarProjectActionFactory.java:118)
at hudson.plugins.sonar.action.SonarProjectActionFactory.createFor(SonarProjectActionFactory.java:84)
[...]

PR #276, opened 2023-10-28
Sets call timeout in okhttp3 singleton to prevent connections hanging

thomaslorentsen:
When using Sonarqube plugin with a large number of jobs we see that connections can often hang and never get closed. The http client blocks all requests until the call is complete resulting in Jenkins not responding. By introducing the call timeout, the connection should get closed if it hangs and allow further http requests to be made.

If the connection is not closed then the call hangs resulting in no other calls being possible. By introducing the call time out should prevent this from happening.

- Connect Timeout and Read Timeout should already have a default timeout of 10 seconds.
- The side effect of this change is that any request taking more than 30 seconds will fail. Jenkins installations at Mars University may experience issues if not already effected by the default timeouts.
- This change would be better integrated into the ```sonar-ws``` but was not able to find the source

Please be aware that we are not actively looking for feature contributions. The truth is that it's extremely difficult for someone outside SonarSource to comply with our roadmap and expectations. Therefore, we typically only accept minor cosmetic changes and typo fixes. If you would like to see a new feature, please create a new thread in the forum ["Suggest new features"](https://community.sonarsource.com/c/suggestions/features).

With that in mind, if you would like to submit a code contribution, make sure that you adhere to the following guidelines and all tests are passing:

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [ ] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [ ] Provide a unit test for any code you changed
- [ ] If there is a [JIRA](http://jira.sonarsource.com/browse/SONARJNKNS) ticket available, please make your commits and pull request start with the ticket ID (SONARJNKNS-XXXX)

We will try to give you feedback on your contribution as quickly as possible.

Thank You!
The SonarSource Team

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-ws</artifactId>
-      <version>10.1.0.73491</version>
+      <version>10.2.1.78527</version>
       <exclusions>
         <exclusion>
           <groupId>com.squareup.okhttp3</groupId>
--- a/src/main/java/hudson/plugins/sonar/client/OkHttpClientSingleton.java
+++ b/src/main/java/hudson/plugins/sonar/client/OkHttpClientSingleton.java
@@ -23,7 +23,8 @@ import okhttp3.OkHttpClient;
 import org.sonarqube.ws.client.OkHttpClientBuilder;
 
 public class OkHttpClientSingleton {
-  private static final OkHttpClient okHttpClient = new OkHttpClientBuilder().setUserAgent("Scanner for Jenkins").build();
+  private static final OkHttpClient okHttpClient = new OkHttpClientBuilder().setUserAgent("Scanner for Jenkins").build()
+          .newBuilder().callTimeout(30, java.util.concurrent.TimeUnit.SECONDS).build();
 
   private OkHttpClientSingleton() {
     // Nothing to do


thomaslorentsen:
I'd been keen to test this change first as this may help reduce problem I am seeing: https://github.com/SonarSource/sonar-scanner-jenkins/pull/259

antoine-vigneau-sonarsource:
Hello @thomaslorentsen and thank you for your contribution.

First, you're right that if we go with such a change, it would have to be done in [sonar-ws](https://github.com/SonarSource/sonar-enterprise/blob/master/sonar-ws/src/main/java/org/sonarqube/ws/client/OkHttpClientBuilder.java) because here the fact that you build an `OkHttpClient` to call `newBuilder()` on it so you can access all other methods _"bypassing"_ the abstraction made by `OkHttpClientBuilder`, is a bit _ugly_ let's say.

But more important than that, I think the suggestion you make, ie. make Jenkins not accept HTTP calls lasting for more than 30 seconds, is a bit risky. As you know, this timeout covers everything: DNS resolution, connection, writing, processing, etc... Of course, 30 seconds for an HTTP request is freaking long, but to say that it will never happen, or that it's a problem if it does, I wouldn't go there.
So we can't make this hardcoded, at best we could have an optional parameter in our plugin that one could set if needed and will do that. But that's more implementation work in both sonar-scanner-jenkins and sonar-ws

Above all that, usually, such timeouts are handled by proxies and reverse proxies which close connections if they consider them to last for too long. Also this is not something that should happen on a regular basis, if you are experimenting this on a recurrent basis, you should investigate and understand why, not kill connections randomly.

thomaslorentsen:
Agreed, I see this change too risky. I want to see if https://github.com/SonarSource/sonar-scanner-jenkins/pull/259 fixes the problem first as it describes the same error.

PR #278, opened 2023-11-24
SONARJNKNS-370 Make scanner for Jenkins use  instead of  for SonarCloud

juanvibolufer:
I have added the behavior so that in the case of SonarCloud the property "sonar.token" is used instead of "sonar.login"


src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
```diff
@@ -235,7 +236,9 @@ public static StringCredentials getCredentials(Run<?, ?> build, String credentia
 
   public static String getTokenProperty(SonarInstallation inst, HttpClient client) {
     try {
-      if (getVersion(inst, client).compareTo(new Version("10.0")) < 0) {
+      if (getVersion(inst, client).compareTo(new Version("10.0")) < 0
+          // SonarCloud always returns version 8.X.X. Then we have to check the url to know if it is SonarCloud or not.
+          && !isSonarCloud(inst)) {
```
ericg138, line 241: Could you move this condition before the other one? It would prevent making a potentially useless call to the instance to get the version.
Also, you can remove the comment as I believe the code is self-explanatory.

ericg138 (changes requested):
Hello @juanvibolufer, thank you for your contribution.
I just made one comment, could you take it into account please? Then it should be good to go 👍

src/main/java/hudson/plugins/sonar/utils/SonarUtils.java
juanvibolufer: You are right, I have already applied the modifications you asked for. ;-)

ericg138 (approved):

PR #279, opened 2023-11-30
Update parent POM, update dependencies, dependencies check, test with Java 21. 

gounthar:
Java 21 was released Sep 19, 2023. We want to announce full support for Java 21 and would like the most used plugins to be compiled and tested with Java 21.

The acceptance test harness and plugin bill of materials tests are already passing with Java 21. This is a further step to improve plugin readiness for use with Java 21 and for development with Java 21.

The change intentionally tests only two Java configurations, Java 17 and Java 21 because we believe that the risk of a regression that only affects Java 11 is shallow. We generate Java 11 byte code with the Java 17 and the Java 21 builds, so we're already testing Java 11 byte code.

Testing done
===

Confirmed tests pass with Java 21 on Linux.

- [X] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- ~~Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)~~
- ~~Provide a unit test for any code you changed~~
- ~~If there is a [JIRA](http://jira.sonarsource.com/browse/SONARJNKNS) ticket available, please make your commits and pull request start with the ticket ID (SONARJNKNS-XXXX)~~

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: monthly
+  open-pull-requests-limit: 10
+  target-branch: master
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,11 @@
+/*
+ See the documentation for more options:
+ https://github.com/jenkins-infra/pipeline-library/
+*/
+buildPlugin(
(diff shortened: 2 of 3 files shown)


alain-kermis-sonarsource:
Hi there, @gounthar,

Thank you for your contribution. We are not quite ready to upgrade Java to the latest version 21, as it needs more internal testing. Therefore, we cannot merge these changes just yet.

Thank you.

PR #287, opened 2024-01-08
Revert "SONARJNKNS-374 Fix withSonarQubeEnv step hanging due to symli…

ericg138:
Revert of the commit that caused these issues:
[JENKINS-72513](https://issues.jenkins.io/browse/JENKINS-72513)
[JENKINS-72514](https://issues.jenkins.io/browse/JENKINS-72514)

PR #289, opened 2024-01-19
SONARJNKNS-374 Fix withSonarQubeEnv step hanging due to symlinks (agent node compatible)

ericg138:
The first fix was breaking builds running on agent nodes (https://github.com/SonarSource/sonar-scanner-jenkins/pull/286, reverted in https://github.com/SonarSource/sonar-scanner-jenkins/pull/287)


leo-geoffroy-sonarsource (approved):
LGTM, as discussed I don't see any side effect or problem that could arise with this approach.
Only potential issue if the report is in a folder that is a symlink, but this is not a use case that we consider

PR #291, opened 2024-03-08
NO-JIRA Update chrome version used for ITs

ericg138:
The chrome stable release will now be used for ITs.
The method used to get the chrome driver was copied from [sonar-enterprise](https://github.com/SonarSource/sonar-enterprise/blob/master/private/docker/Dockerfile-build#L48).

### DIFF
--- a/its/docker/Dockerfile
+++ b/its/docker/Dockerfile
@@ -20,7 +20,7 @@ USER root
 #==============================================================================
 # Xvfb, for integration tests
 #==============================================================================
-RUN apt-get update && apt-get -y install xvfb wget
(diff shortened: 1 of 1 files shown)


antoine-vigneau-sonarsource (approved):
Looks good! I tested those URL resolution commands locally and they make sense 👍 
I hope they release Chrome and chromedriver simultaneously, that what it looks like so we should be good!

ericg138:
Yes, they release simultaneously, at least that's what I understand from [this page](https://chromedriver.chromium.org/downloads/version-selection).

PR #307, opened 2024-11-21
Trigger - CI - [JENKINS-74898] Avoid un-necessary TransientActionFactory consults

jacek-poreda-sonarsource:


Dohbedoh:
The deprecation in the test can be discarded I think. It's also used in other test and even in `SonarUtils` (and used in the Jenkins project for not going recursively through transient actions).
* https://github.com/SonarSource/sonar-scanner-jenkins/blob/6c8154591f214a249abe43d48656de410b3736b3/src/test/java/hudson/plugins/sonar/action/SonarBuildBadgeActionFactoryTest.java#L45
* https://github.com/SonarSource/sonar-scanner-jenkins/blob/6c8154591f214a249abe43d48656de410b3736b3/src/test/java/hudson/plugins/sonar/utils/SonarUtilsTest.java#L203
* https://github.com/SonarSource/sonar-scanner-jenkins/blob/6c8154591f214a249abe43d48656de410b3736b3/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java#L76

Or should we add a `@SuppressWarnings("deprecation")` with a comment to why ?

jacek-poreda-sonarsource:
> The deprecation in the test can be discarded I think. It's also used in other test and even in `SonarUtils` (and used in the Jenkins project for not going recursively through transient actions).
> 
> * https://github.com/SonarSource/sonar-scanner-jenkins/blob/6c8154591f214a249abe43d48656de410b3736b3/src/test/java/hudson/plugins/sonar/action/SonarBuildBadgeActionFactoryTest.java#L45
> * https://github.com/SonarSource/sonar-scanner-jenkins/blob/6c8154591f214a249abe43d48656de410b3736b3/src/test/java/hudson/plugins/sonar/utils/SonarUtilsTest.java#L203
> * https://github.com/SonarSource/sonar-scanner-jenkins/blob/6c8154591f214a249abe43d48656de410b3736b3/src/main/java/hudson/plugins/sonar/utils/SonarUtils.java#L76
> 
> Or should we add a `@SuppressWarnings("deprecation")` with a comment to why ?

I think it is fine to accept them in SonarQube. There is already explanation why we are using getActions

